Issue #539, opened 2015-01-08
Missing useragent strings in 5037


ericberdahl:
I've encountered the following user agent strings that are not diagnosable with version 5037 of full_asp_browscap.ini

```
(DreamPassport/3.0; isao/MyDiGiRabi)
-
ALCATEL-OT871A/980 Profile/MIDP-2.0 Configuration/CLDC-1.1 ObigoInternetBrowser/Q05A
Android
Android App Ver 3.0.3 (Android 2.3.3;)
BTWebClient/3420(35702)
CookieReports.com/0.1
Desktop
DumontMobileWrapperAndroid
EMail Exractor
Elmer, the Thinglink ImageBot (http://www.thinglink.com/help/ThinglinkImageBot)
FlixsteriOS/6.15 (iPad; CPU OS 7_0 like Mac OS X) AppleWebKit/537.51.1 (KHTML, like Gecko) Mobile/11A465
FlixsteriOS/6.18.4 (iPad; CPU OS 7_0_4 like Mac OS X) AppleWebKit/537.51.1 (KHTML, like Gecko) Mobile/11B554a
HTC_ChaCha_A810e/Mozilla/5.0(Linux; U; Android 2.3.5;it-it; Build/GRH78C) AppleWebkit/533.1 (KHTML, like Gecko) Version/4.0 Mobile Safari/533.1
Iframely/0.7.1 (+http://iframe.ly/;)
Lenovo A789/S210 Linux/3.0.13 Android/4.0 Release/05.11.2012 Browser/AppleWebKit533.1 Profile/ Configuration/
Lenovo-A880/S100 Linux/3.4.5 Android/4.2 Release/08.07.2013 Browser/AppleWebKit 534.30 Profile/ Configuration;
MQQBrowser/2.7 Mozilla/5.0 (Linux; U; Android A5000; in-id; A5000 Build/GRK39F) AppleWebKit/533.1 (KHTML, like Gecko) Version/4.0 Mobile Safari/533.1
MQQBrowser/36 Mozilla/5.0 (iPhone 4S; CPU iPhone OS 5_1_1 like Mac OS X) AppleWebKit/534.46 (KHTML, like Gecko) Mobile/9B206 Safari/7534.48.3
MQQBrowser/Mini2.4 (SAMSUNG-GT-E2252)
MSIE 6.0
Mass Times
MetaInspector/3.0.0 (+https://github.com/jaimeiniesta/metainspector)
Microsoft-CryptoAPI/6.1
[...]

PR #545, opened 2015-01-24
Issue 539

mimmi20:
I added the most UAs mentioned in #539 and fixed the detection for them. 

For this I needed to add a new Devicetype.

This should solve the Issue.


asgrim:
I think this needs an update from master :)

mimmi20:
Update done

Issue #555, opened 2015-02-08
New User-Agent bot@qwant.com


hcoding:
browscap.ini Version 5038

Using:
browscap/browscap-php

User-Agent:
Mozilla/5.0 (compatible; Qwantify/2.0; +https://www.qwant.com/)

PR #556, opened 2015-02-09
Issue 555

mimmi20:
I added the new rule and a test. This should solve Issue #555.


asgrim:
Hmm, I don't see any reason why the hhvm build is failing. I'll merge this but we can look into the hhvm build separately. Thanks @mimmi20  :)

PR #557, opened 2015-02-10
avoid error during build

mimmi20:
This PR  should prevent build errors when the temporary build dire already exists.


asgrim:
Thanks @mimmi20 :)

Issue #558, opened 2015-02-11
Lenovo A7600 Tablet Issue. Not picked up as a Tablet


torches:
Hi,

We have a tablet that is not returning true for isTablet.
The user agent is: Mozilla/5.0 (Linux; Android 4.4.2; Lenovo A7600-F Build/KOT49H) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/40.0.2214.109 Safari/537.36

Would you mind adding this to your DB.
I checked first on the UA tester, and the latest DB does not work with it.

Thanks

PR #559, opened 2015-02-11
Issue 558

mimmi20:
I added new platform versions for the device and a test. This should solve Issue #558.


asgrim:
Thanks @mimmi20  :)

Issue #491, opened 2014-10-26
extract device properties


mimmi20:
To reduce duplicate content inside the resources json files, all device related properties should be extracted into a separate file. This should be done in the same way the engines and platforms were extracted before.

PR #564, opened 2015-02-25
Issue 491

mimmi20:
I have extracted the device information from the useragents and the platform definitions. This was made to have a central location where the devices with their properties are defined to prevent misconfigurations due multiple usages of the device.

This should solve Issue #491.


asgrim:
Nice :)

PR #614, opened 2015-04-08
Reduce the size of the "lite" files significantly

### LINKED ISSUE #543: Reduce size of lite file further
I'd also like to further review the lite browsers to ensure only modern browsers/crawlers are included. How should we reliably determine this? How do we review this in future?

asgrim:
I have examined the issue causing such large files, and it seems to come from the fact there are so many platform variations that causes exponential file size increase. This PR takes two steps that enormously reduce the file size (and in fact, I think perhaps it may have been *too* aggressive and we can probably increase the file size a bit):

* Removed the "lite" flag from a lot of browsers - some are "modern" but mobile only, some where truly unworthy of being included in the lite files. I think I've been a little aggressive on this part, so it may be worth reviewing all the browsers to see what we could add in to make a little more comprehensive file, without sacrificing too much filesize.
* Added a way to exclude entire platforms from the output (in lite mode) - now there is only "general" detection (e.g. we can detect `*Mac OS X*` but not the specific version e.g. `*Mac OS X 10_4*` in the lite file). This should allow sufficient platform detection for lite versions, but without the extra detail that the specific platform versions give. In doing so I've added the `isOutputSection()` method to the `FilterInterface` that checks the "lite" flag.

You can try out the lite versions on these temporary URLs:
* http://browscap.org/6002-lite-test/lite_php_browscap.ini
* http://browscap.org/6002-lite-test/lite_asp_browscap.ini

Comments and suggestions are welcome. Merging this would resolve #543


asgrim:
@mimmi20 comments / suggestions welcome on this PR! :)

asgrim:
You can try out the lite versions on these temporary URLs:
* http://browscap.org/6002-lite-test/lite_php_browscap.ini
* http://browscap.org/6002-lite-test/lite_asp_browscap.ini

mimmi20:
@asgrim I tried some weeks ago to add tests for the lite and the standard mode. But that failed on the 10 minutes limit on travis-ci.org. Because of the lot of issues poped up then I did not search any further.

It would be nice if you would create unittests to cover you add or change.

By looking into the created lite file I have the following comments:
- All Platforms/Platform Versions with 64 Bit (like wow64, win64, x86_64) and some with 32 Bit (like i686) can be removed from the lite file, because the Bits information is not available in the lite file and these rules are covered by less specific rules.
- Some rules for the iPod are not needed in the lite file. With these rules there are not additional data in the lite file. But I dont know if this can be removed easely.
- In the created lite file there are some sections like `Safari * for SymbianOS`. These sections are not used anywhere, because SymbianOS is removed from the lite file. Could you check, if you can remove theses sections too? There are also unused sections for Windows RT or Maemo.
- The last section `[*]` (the Defaultbrowser) is missing.

asgrim:
For my own reference:  `hhvm -c "" bin/browscap build xxx`

asgrim:
@mimmi20 the Default Browser section is also missing in the current lite file (!) - it was reported as bug in #620 so we'll solve that separately as this was not broken because of this PR

asgrim:
@mimmi20 addressing your comments:

* I've removed most of them now, the rest should be negligible now
* I think I managed to remove some iOS defintiions that were not required (I left `iOS_C` as it was the most generic)
* I'm not sure how easy it would be to remove those, so I have left in for now - they are child definitions and my attempts to get them ignored have failed
* See my previous comment about the `[*]` section (reported in #620, solve separately)

I did add a few more UAs back in (specifically, IE6/7 and Chromium). Is there any others you think we can add back in perhaps? I think the biggest "bang for buck" was removing the many platform definitions, so we could consider adding some UAs back in now.

mimmi20:
@asgrim I could not add tests for funtions you added in this branch like `isOutputSection` for the writers. It would be nice I you add some tests to cover these functions.

asgrim:
Cool - will do :)

asgrim:
@mimmi20 tests added, so coverage remains the same after the latest Scrutinizer check. Any other feedback please? Thanks!

mimmi20:
@asgrim I have no other comments.

asgrim:
Thanks :)

Issue #633, opened 2015-04-24
Project Spartan Incorrectly Reports ActiveX Support


jimstump:
Using the latest (6001) full version of Browscap, both on my server using GaretJax/phpbrowscap and using the UA lookup tool on browscap.org, fetching information about Project Spartan shows support for ActiveX, which is not currently supported.  I know that Project Spartan is still in development, but from what I have read around the Internet, support in Spartan directly seems unlikely.

User Agent is:

Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.71 Safari/537.36 Edge/12.0 

I would expect the ActiveXControls property to be false for Project Spartan.  Of course, IE <= 11 does support ActiveX.

PR #643, opened 2015-05-06
Issue 633

mimmi20:
I renamed `IE 12` to `Edge` and added the `Edge` engine.

For the other changes I suppose the following things.
- our `BackgroundSounds` property is related to the `bgsound` HTML attribute which is supported by IE (Trident) only
- our `VBScript` property is related to VBScript which is supported by IE (Trident) only
- our `ActiveXControls` property is related to ActiveX which is supported by IE (Trident) only
- our `JavaApplets` property is related to the `applet` HTML element which is supported by IE (Trident), Firefox (Gecko) and Safari (WebKit), but not in Chrome and Opera (I do not remember how old the site with these information was)

For easier handling I moved these properties into the engines.json file and removed them inside all useragent json files.

@asgrim To enable the `JavaApplets` for Safari I swiched the property on for WebKit and disabled the property for Chrome. I do not know if Gecko actual supports this Element. Should I disable this property for WebKit and Gecko too? This HTML element was never part of a HTML specification and should not be used with HTML5.

All these four properties are disabled for the new Edge engine. This PR should solve Issue #633.


asgrim:
@mimmi20 [this link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/applet) indicates that it works in Gecko but is obsolete.

asgrim:
so yeah - happy to accept this if you are?

mimmi20:
I'm fine with this.

Issue #651, opened 2015-05-15
"CFNetwork" Entries


tjasond:
There appears to be a few instances of miscategorization for user agents containing "CFNetwork".   The most eager of these is in the Image Crawlers section, where *CFNetwork* will match all variations of this user agent.  Also, the "CFNetwork/*" expression in the Rippers section will match all of them as well.

From what I'm reading, all iOS apps that use the built in networking API (called CFNetwork) will use some form of this expression in the user agent by default:
http://stackoverflow.com/a/8031061

Some of these are crawlers, like this one:
Tweetbot/* CFNetwork/* *Darwin*

although others are legitimate apps, such as the Flipboard app that is specified separately:
Flipboard/1.5* CFNetwork/* *Darwin*

and there's evidence that this user agent in the Googlebot section is actually the Google App for iOS:
Google/1.0* CFNetwork/* *Darwin*
https://productforums.google.com/forum/#!topic/webmasters/-dBb6GLiSM8

If this has been previously addressed, please disregard - this is in reference to version 6001 of the full_asp_browscap.ini file.

Thanks

mimmi20:
@tjasond Could you add some useragents for Apps or Browsers which should descriped a better way?

mentor2k1:
Hi,

i´ve started more than one year ago to investigate the CFNetwork UA.

Here you get a nearly complete cfnetwork version list: http://user-agents.me/cfnetwork-version-list

and Some App UA Strings: http://user-agents.me/search?q=cfnetwork

// Oliver

tjasond:
@mimmi20 Thanks for following up.  @mentor2k1 links to a pretty comprehensive list (thanks by the way), although my concern is more focused on the possible miscategorization of all user agents containing "CFNetwork" being designated as a Bot/Crawler.  If the two "CFNetwork" expressions (in the Rippers and Image Crawlers sections) that I referenced earlier were removed, as well as the Google CFNetwork entry recategorized as the Google App and not a bot, then that would address my concern.  

As far as whether or not to include/maintain all of the possible CFNetwork user agent variations, my opinion would be to only include the noteworthy ones such as bots/crawlers, or popular apps such as the Google App, which appears to be the direction taken up to this point.  If you attempt to classify and maintain every CFNetwork variation, it will be a very high cost, and the value, at least in my use cases, would be disproportionate to the cost.  

Thanks again.

mentor2k1:
[...]

PR #656, opened 2015-05-25
Issue 651

mimmi20:
I moved some rules from browsers/crawlers to apps, added new versions and rules. This should solve Issue #651.


tjasond:
Thanks for fixing this, works as expected now in 6003.

Issue #620, opened 2015-04-15
Lack of default in PHP LITE file


addw:
The other 2 have a line near the end:
[*]
This is useful not just as a default but also checking (via script) that the download has completed fully.
Please can we have this, or some other end-of-file marker that can be checked.
Thanks

asgrim:
Which version of the file are you using?

addw:
I downloaded  http://browscap.org/stream?q=Lite_PHP_BrowsCapINI a few minutes ago, size is 8744244 and the last entry is
 [Mozilla/5.0 (*Android*Tablet*)*Gecko/*Firefox/ anonymized by Abelssoft*]

asgrim:
6001

addw:
Yes - 6001

asgrim:
Re-opening to check for regressions - I am adding unit test to ensure the * section is present when generating the files.

asgrim:
Need #617 to be merged so I can work on this as a base

PR #661, opened 2015-06-01
Add regression test to ensure Default Browser always exists

asgrim:
This adds a regression test to ensure #620 doesn't break again (i.e. all browscap files should contain the `Default Browser` entry, including lite).


asgrim:
@mimmi20 please review this little one when you get a sec :)

mimmi20:
@asgrim This looks fine to me.

Issue #699, opened 2015-07-16
Last versions of Opera Mini are not detecting (v 10)


velanse:
The following user agents are Opera Mini v 10.2:

- Mozilla/5.0 (iPhone; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) OPiOS/10.2.0.93022 Mobile/12F70 Safari/9537.53

- Mozilla/5.0 (iPhone; CPU iPhone OS 8_4 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) 
OPiOS/10.2.0.93022 Mobile/12H143 Safari/9537.53

But they are detecting as Safari Browser via browscap.

Issue #698, opened 2015-07-16
Windows 10 Edge browser incorrectly recognized as Chrome


fberzau:
This user agent string from the most current Windows 10 build recognizes the Edge browser as Chrome:

Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.135 Safari/537.36 Edge/12.10240

magorinahory:
Hi, this issue should be addressed in 6006.
Please review this closed issue https://github.com/browscap/browscap/issues/688
and code https://github.com/browscap/browscap/commit/815b20f9a28fb56e2b9e8612c33d85131f39a5f5

Issue #697, opened 2015-07-15
tablet detection


jota78:
Both of these browsers are for Tablets but these are being detected as isTablet:False

Mozilla/5.0 (Linux; Android 4.1.2; GT-P3100 Build/JZO54K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/43.0.2357.93 Safari/537.36

Mozilla/5.0 (Linux; U; Android 4.1.2; en-us; E-TAB IVORY Build/E702) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 Safari/534.30

There are probably a lot more like these.

PR #700, opened 2015-07-16
Issue 699

mimmi20:
I added new Versions for iOS and Opera Mini. I also added a new Device, new rules and tests.

This should solve the Issues #697, #698 and #699.


velanse:
Thank you, Thomas!

PR #734, opened 2015-09-08
updating tests to new master

mimmi20:
This PR will change tests to use the new master of browscap/browscap-php.

Actually builds on hhvm will fail because the job is killed. I asked Travis about that. They said the memory limit is reached. On a local hhvm instance the tests were finished successful. 

One test is marked as skipped because it fails on PHP 7 when creating the zip file.


asgrim:
Hmm - can't merge this directly yet, as it impacts browscap/browscap-site#17 which is not ready.

Also - memory limit *surely* should not be a problem? in 3.x branch, the memory usage dropped really low (at the expense of performance) - see browscap/browscap-php#101

mimmi20:
I could not verify this, on my local HHVM instance it worked. But every test fails for HHVM because the process is killed.

asgrim:
@mimmi20 can you also squash the merge commits out here please?

mimmi20:
I am not able to squash the merges out. I will clode this PR and open a new one.

Issue #761, opened 2015-09-29
full_php_browscap.ini (version 6007) has syntax errors


cyrax-02:
When i run this code:
get_browser();

i have 20-40 warning messages:
PHP warning: get_browser() [<a href='function.get-browser'>function.get-browser</a>]: Compilation failed: missing terminating ] for character class at offset 21

it seems that new (full_php_browscap.ini) has syntax errors.
In preview version (6006) these warnings absent.

asgrim:
I will release 6008 tomorrow hopefully to fix this.

PR #769, opened 2015-10-04
add new check

mimmi20:
This PR adds a new check for characters in the rules with the possibility to create issues.

This PR is related to Issue #761.


src/Browscap/Data/DataCollection.php
```diff
@@ -252,6 +252,16 @@ public function addSourceFile($src)
 
         if (isset($divisionData['userAgents']) && is_array($divisionData['userAgents'])) {
             foreach ($divisionData['userAgents'] as $useragent) {
+                if (!isset($useragent['userAgent'])) {
+                    throw new \UnexpectedValueException('Nsme for Division is missing');
```
asgrim: typo in `Name`

mimmi20: is fixed

Issue #750, opened 2015-09-24
iOS 9 Apps using CFNetwork


mimmi20:
see ua-parser/uap-core#89 for more details

PR #799, opened 2015-10-23
Issue 750

mimmi20:
I added two new tests and a map from `CFNetwork` versions to `iOS`/`MacOSX` versions for all browsers/applications we are testing (All other browsers/applications should be changed with seperate tickets).
This should solve Issue #750.


asgrim:
thanks @mimmi20

Issue #808, opened 2015-10-28
User Agents with specific wrong strings


mapanet:
Will be helpful to include some UA strings that are specific and obviously wrong thing.
Return as "Default Browser" is enough to know is not a legit browser/connection UA.
Like this, where the whole UA string is:

Googlebot
test certificate info

asgrim:
The UA `test certificate info` returns Default Browser. `Googlebot` is matched by `googlebot*` to ensure reasonable forwards compatibility. This is essentially a duplicate of #3, so closing this in favour of that.

Issue #806, opened 2015-10-27
New User Agent (Crawler)


mapanet:
Mozilla/5.0 (compatible; DeuSu/0.1.0; +https://deusu.de/robot.html)
Mozilla/5.0 (compatible; DeuSu/5.0.2; +https://deusu.de/robot.html)

MichaelSchoebel:
Do you need any info about this bot? I'm the person behind it...

The crawled data is used for two websites. https://deusu.de (German) and https://deusu.org (English).

And BTW: Sources for the software used for this bot are on
https://github.com/MichaelSchoebel/DeuSu

asgrim:
Thanks @MichaelSchoebel for contact us - might be worth checking https://github.com/mimmi20/browscap/blob/0ba9c07cb1b39be6413856a754ea5f306c2842af/tests/fixtures/issues/issue-806.php this file for validity! Otherwise, all good thanks :+1:

mapanet:
Congrats Michael, thumbs up!

Checked the page before submitting a browscap.org request because noticed
it was a real bot, so I asked to add/review properties since had almost
none at all, specially crawler=*false*, a very important property in my
opinion. Also caught my attention a version 0.1.0 jumping to 5.0.2 in a few
days which triggered my curiosity.

Maybe will be to much bla, bla, but since you asked... ;-)

As many others, I use keywords for User-Agent checking for good and bad
bots, *deusu* string was caught by my unknown / bads.
I research the unknowns to figure where they belong: good, bad & beyond
(Chinese), and position it earlier in the array if comes back often.
deusu.de and .org are welcome since yesterday, thanks to your page which
identifies clearly, says what is all about, and more important every day,
how collected data will be used:

   - Do they will sell the data collected?
   - Do they get data to offer it to the competitors?
   - Do they publish stats of sales of my website, how do they know? ;-)
   - Do they make business with keywords found for SEO business ?

Today we know why, 70% of traffic is bots, spiders, scrappers, Chinese,
Russian, spies (oh, that's Chinese too), 24 hours straight drilling
websites seem to be fun for many. So the more details, the easier is
accepted, specially in near future when the new W3C rules and features for
user agents finally arrive.

Basic properties play a big role, example: I use session variables so have
a 20 minutes session.timeout, by knowing is a bot, I can decrease
session.timeout to 1 minute, lowering IIS server load. This showed benefits
in just two weeks, 3000 active sessions to 700. Since bot visits does not
[...]

Issue #805, opened 2015-10-27
Missing User Agent


mylearningcenter:
browsecap seems to be missing the following user agent:

Mozilla/5.0 (Windows NT 6.1; WOW64; Trident/7.0; BOIE9;ENGB; rv:11.0) like Gecko

Windows 7, IE 11

asgrim:
@mylearningcenter please try the latest version (6009) - I looked that UA up on http://browscap.org/ua-lookup and it worked just fine...

mylearningcenter:
Thanks.  Maybe I'm doing something wrong then because I was using v6008 but I just downloaded v6009 as you requested and I still don't see it.  I was using the browscap.csv file and I assume there aren't any differences between the file types as far as content.  As a simple test, I tried grepping BOIE9 in that CSV and found no matches so I was wondering how the ua-lookup used the files.  Maybe I'm using the CSV incorrectly.

Thank you.

asgrim:
The CSV cannot be used directly, it is a CSV that contains "matching" patterns, that have tokens like `*`, i.e.:

```
Mozilla/5.0 (*Windows NT 6.1*WOW64*Trident/7.0*rv:11.0*)*
```

This pattern is associated to the metadata that can be seen on http://browscap.org/ua-lookup

You cannot search directly for a UA in any of the files we provide, you need to use a library to parse these (and most likely cache the matching patterns for performance) and look up the UA this way.

See: https://github.com/browscap/browscap/wiki/Using-Browscap for help

mylearningcenter:
Thanks.  That explains it.  As a suggestion, it might be a good idea to post the wiki link on the main site page because I looked for some documentation and didn't see it right off.
Thanks for the help.

asgrim:
@mylearningcenter thanks, will do :)

asgrim:
@mylearningcenter FYI, this has been implemented ;)

PR #813, opened 2015-10-29
Issue 806

mimmi20:
I added the new Bot and some tests. This should solve Issue #806.

I added some tests for the Uas from the Issues #805 and #808 too.


asgrim:
thanks @mimmi20  :)

Issue #811, opened 2015-10-28
some more missing UA


gocher68:
Mozilla/5.0 (compatible; Scopia Crawler 1.2; +http://www.scopia.co)
Mozilla/5.0 (Windows NT 6.3; Win64; x64; rv:26.0.0b2) Goanna/20150828 Gecko/20100101 AppleWebKit/601.1.37 (KHTML, like Gecko) Version/9.0 Safari/601.1.37 PaleMoon/26.0.0b2
Opera/9.80 (Window NT 6.2; Win64; x64) Presto/2.12.388 Version/12.16
Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:43.0) Gecko/20100101 Firefox/43.0
Mozilla/5.0 (Mobile; rv:32.0) Gecko/20100101 Firefox/32.0
Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:44.0) Gecko/20100101 Firefox/44.0
Mozilla/5.0 (Windows NT 10.0; WOW64; rv:43.0) Gecko/20100101 Firefox/43.0
Mozilla/5.0 (Windows NT 7.0; WOW64; rv:40.0) Gecko/20100101 Firefox/40.0
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_4 AppleWebKit/537.36 (KHTML, like Gecko) Chrome/36.0.1985.125 Safari/537.36
Mozilla/4.0 (compatible; MSIE 5.5;Windows NT)
Mozilla/5.0 (Windows NT 6.2; Win64; x64; rv:16.0) KcB/16.0 Firefox/16.0
Mozilla/5.0 (Unknown; Linux x86_64) AppleWebKit/538.1 (KHTML, like Gecko) PhantomJS/2.0.1-development Safari/538.1

mimmi20:
Does anyone know, what the `KcB/16.0` Token in the second last Ua is for?

PR #774, opened 2015-10-07
Firefox Nightly is in version 44

Zefling:

PR #818, opened 2015-10-31
Issue 811

mimmi20:
I added two new Bots, a new engine and versions for FirefoxOS, added new versions to PaleMoon, Firefox and PhantomJS.

This shuld solve Issue #811. This also makes PR #774 obsolete.


asgrim:
thanks @mimmi20

PR #831, opened 2015-11-12
Issue 542

mimmi20:


mimmi20:
Needs some more checks

mimmi20:
@asgrim Can we replace the `Win16`, `Win32` and `Win64` properties with the `Crawler` property in standard mode? This would help to reduce the file size.

asgrim:
Yes :) don't see a problem with that

mimmi20:
All Issues I found are fixed now. Plaese merge PR #834 before this one.

Issue #849, opened 2015-11-24
examples of user agent strings


mimmi20:
See ua-parser/uap-core#117 for more details.

PR #854, opened 2015-11-29
Issue 849

mimmi20:
I added some new bots and tests. I also re-grouped the rippers and download managers a bit. This should solve Issue #849.


asgrim:
@mimmi20 thanks

Issue #855, opened 2015-11-29
incorrect display for User Agent


mimmi20:
moved from browscap/browscap-php#128.

Hello,
When I go to the browser "Google Chrome", UA Lookup wrote that I went to the browser chrome.
But if to come to the browser "YandexBrowser", UA Lookup wrote that I went to the browser chrome, instead of to write browser "YandexBrowser".

Когда я захожу с браузера "Google Chrome", UA Lookup пишет что я зашел с browser chrome.
Но если зайти с браузера "YandexBrowser", UA Lookup пишет что я зашел с browser chrome , вместо того чтобы написать browser "YandexBrowser".

User Agent:
-YandexBrowser:
Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.101 YaBrowser/15.10.2454.3865 Safari/537.36

-Google Chrome:
Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.86 Safari/537.36

PR #856, opened 2015-11-29
Issue 855

mimmi20:
I added some new browser versions and platforms to the Yandex Browser. This should solve Issue #855.


asgrim:
@mimmi20 thanks

PR #864, opened 2015-12-15
Generate a lite_browscap.json

lwille:
The full browscap.json is too big for our project - it consumes about 1.2 GB of process memory in our node.js browscap microservice.

The lite build is (surprise!) much more lightweight, it only consumes ~50 MB.

Please also add the lite_browscap.json to the Downloads page after accepting this pull request.

### DIFF
--- a/src/Browscap/Writer/Factory/FullCollectionFactory.php
+++ b/src/Browscap/Writer/Factory/FullCollectionFactory.php
@@ -136,6 +136,15 @@ class FullCollectionFactory
         ;
         $writerCollection->addWriter($jsonWriter);
 
+        $liteJsonWriter = new JsonWriter($buildFolder . '/lite_browscap.json');
+        $formatter  = new JsonFormatter();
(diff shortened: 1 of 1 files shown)


asgrim:
@mimmi20 review please?

lwille:
ARgh, I just noticed that the format seems to be different (version information is only provided as a "Version" property instead of "MajorVer" and "MinorVer"), so it's not really compatible with the full version :cry:

asgrim:
@lwille indeed, we strip out a lot of cruft in the lite version to reduce file size... can you not just use `Version`?

lwille:
Sure, but there are some other capabilities that our legacy app is interested in (like JavaScript).

I think I'll just do some filtering when loading the full data to keep only the properties I need.

I'll leave this pull request, maybe someone else is interested in the vanilla lite format :)

mimmi20:
The changes look OK to me.

Issue #873, opened 2015-12-26
Missing browsers


mimmi20:
see bestiejs/platform.js#82 for more details.

PR #874, opened 2015-12-26
Issue 873

mimmi20:
I added a new device, some new platform versions, new rules and two new tests. This should solve Issue #873.


asgrim:
thanks @mimmi20

Issue #893, opened 2016-02-04
Errors in 6012 build


asgrim:
https://gist.github.com/asgrim/c8ae1a8b2a4c4934f56e

PR #896, opened 2016-02-12
Issue 893

mimmi20:
I fixed the definition for IE 7.0b to solve the duplicate enties during build. This should solve Issue #893.


asgrim:
thanks @mimmi20

Issue #902, opened 2016-02-18
Add Outlook 2016 Support


mimmi20:
see ua-parser/uap-core#132 for more details

PR #903, opened 2016-02-21
Issue 902

mimmi20:
I added 3 tests and splitted Office 2016 from Outlook (like the other Office versions). This should solve Issue #902.


asgrim:
Not sure why the PR status hasn't been updated here, but it looks like tests pass :s

Issue #905, opened 2016-03-04
Unknow Lynx


mimmi20:
see woothee/woothee#22 for more details

asgrim:
Closed by #910

PR #910, opened 2016-03-11
Issue 905

mimmi20:
I added a new test. This should solve Issue #905.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "source": "https://github.com/browscap/browscap"
   },
   "require": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -31,7 +31,7 @@
     "source": "https://github.com/browscap/browscap"
   },
   "require": {
-    "php": ">= 5.5.0",
+    "php": ">= 5.6.0",
```
asgrim: The change to min version should ideally be done in a separate PR but I'll accept in this case :|

Issue #941, opened 2016-04-07
Nexus 7 running Androïd 6.0.1 not detected


effetb:
Hi,

I'm using browscap with php get_browser function, and my nexus 7 is not detected anymore since I updated it to 6.0.1
I looked at the browscap.ini file (updated every day version = 6013 ) and didn't found any 6.x android version in it, is this normal ?

The user agent string from my Nexus 7 :
Mozilla/5.0 (Linux; Android 6.0.1; Nexus 7 Build/MMB29V) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/49.0.2623.105 Safari/537.36

Regards

PR #944, opened 2016-04-07
Issue 941

mimmi20:
I added a new versions to Chrome and splitted the source files. I added the new platform version to the newest versions of Chrome. This should solve Issue #941.


asgrim:
Thanks!

Issue #920, opened 2016-03-22
ucbrowser detection


pascalsiegel:
Hello,

it seems UC browser is not well detected, below are the UA and what is returned for the browser name and version:

On windows 7, please not the spelling in the UA "UBrowser"
UA : Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/48.0.2564.116 UBrowser/5.6.10764.1203 Safari/537.36
name : Chrome
version : 48.0

This one if from UC Browser HD on an Android 4.2.2 tablet! please not that it is returning iPad!!!
UA: Mozilla/5.0 (iPad; U; CPU OS 5_1 like Mac OS X) AppleWebKit/531.21.10 (KHTML, like Gecko) Version/4.0.4 Mobile/7B367 Safari/531.21.10 UCBrowser/3.4.3.532
name : Safari
version : 0.0

From the same tablet but with UC Browser (not HD version)
UA: Mozilla/5.0 (Linux; U; Android 4.2.2; en-US; IdealTab 10 Build/JDQ39) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 UCBrowser/10.8.8.730 U3/0.8.0 Mobile Safari/534.30
name : Android
version : 4.0

PR #945, opened 2016-04-07
Issue 920

mimmi20:
I added a new device and the new browser version and three tests. This should solve Issue #920.


asgrim:
Thanks!

Issue #904, opened 2016-02-21
Detection Issues


mimmi20:
There are many detection issues highlighted in parser comparasation ThaDafinser/UserAgentParserComparison.

See http://thadafinser.github.io/UserAgentParserComparison/v4/ for more details.

PR #963, opened 2016-05-02
Detection Issues

mimmi20:
I fixed some issues found with ThaDafinser/UserAgentParserComparison and added some bots. This PR is related to Issue #904.


.php_cs
```diff
@@ -0,0 +1,95 @@
+<?php
```
asgrim: If this `.php_cs` is intentionally added, can you add it in a separate PR please?

mimmi20: I removed the file

PR #1004, opened 2016-05-29
change data providers for unittesting

mimmi20:
With PR the PHPUnit configuration is splitted into two files, one for the tests on this project and one for the comparasation tests.

The reason is that the dataproviders of the comparasation tests are also executed in the project tests, even when tests are ignored via exclude-group.

I also changed the data providers for the comparasation tests a bit.


asgrim:
One of the files should simply be called `phpunit.xml`, so that when someone runs `vendor/bin/phpunit` they get the basic test suite.

What's the reason `--exclude-group` isn't working though?

mimmi20:
The phpunit.xml in the base path was not removed. The `exclude-group` was working also. The issue (in my mind) was that the data providers were execuded before filtering the tests. I thought test classes are skipped when the class itself has a `@group` signature with the group to skipped.
Because the performance gain is really low, I removed the new phpunit files. Only changes in the data providers are left.

mimmi20:
@asgrim Please review this again.

tests/BrowscapTest/UserAgentsTest.php
```diff
@@ -142,21 +141,82 @@ public function userAgentDataProvider()
     }
 
     /**
-     * @dataProvider userAgentDataProvider
+     * @return array[]
+     */
+    public function userAgentDataProviderFull()
+    {
+        $providerData = array();
```
asgrim: `$providerData = [];`

tests/BrowscapTest/UserAgentsTest.php
```diff
@@ -142,21 +141,82 @@ public function userAgentDataProvider()
...
+
+        foreach ($this->userAgentDataProvider() as $key => $test) {
```
asgrim: wouldn't this be more efficient with an `array_filter`?

tests/BrowscapTest/UserAgentsTest.php
```diff
@@ -142,21 +141,82 @@ public function userAgentDataProvider()
...
+    }
+
+    /**
+     * @return array[]
+     */
+    public function userAgentDataProviderStandard()
+    {
+        $providerData = array();
+
+        foreach ($this->userAgentDataProvider() as $key => $test) {
```
asgrim: wouldn't this be more efficient with an `array_filter`?

tests/BrowscapTest/UserAgentsTest.php
```diff
@@ -142,21 +141,82 @@ public function userAgentDataProvider()
...
+    }
+
+    /**
+     * @return array[]
+     */
+    public function userAgentDataProviderLite()
+    {
+        $providerData = array();
+
+        foreach ($this->userAgentDataProvider() as $key => $test) {
```
asgrim: wouldn't this be more efficient with an `array_filter`?

mimmi20: changed

tests/BrowscapTest/UserAgentsTest.php
mimmi20: changed

tests/BrowscapTest/UserAgentsTest.php
mimmi20: changed

tests/BrowscapTest/UserAgentsTest.php
mimmi20: fixed

Issue #993, opened 2016-05-15
Podcast catcher Applications


mimmi20:
see ua-parser/uap-core#148 for more details

PR #1007, opened 2016-06-05
remove platform properties from useragents-section

mimmi20:
With this PR the properties `Win16`, `Win32`, `Win64` and `Browser_Bits` are removed from the resources within the user-agents directory.
The reason is that these properties are defined in the `platforms.json` file. These properties inside the user-agents directory are ignoed (when not changed) or overwritten.


mimmi20:
@asgrim Could you merge this PR? I need this to finish issue #993.

asgrim:
Thanks!

PR #1019, opened 2016-07-07
Issue 1017

mimmi20:
I fixed the failing test and rewrote some unittests to ger rid of the warings.

### DIFF
--- a/tests/BrowscapTest/Data/DataCollectionTest.php
+++ b/tests/BrowscapTest/Data/DataCollectionTest.php
@@ -82,7 +82,9 @@ class DataCollectionTest extends \PHPUnit_Framework_TestCase
      */
     public function testSetGetLogger()
     {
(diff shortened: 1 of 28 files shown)


tests/BrowscapTest/Data/DataCollectionTest.php
```diff
@@ -82,7 +82,9 @@ private function getUserAgentFixtures()
      */
     public function testSetGetLogger()
     {
-        $logger = $this->getMock('\Monolog\Logger', [], [], '', false);
+        $logger = $this->getMockBuilder(\Monolog\Logger::class)
```
asgrim: `createMock` should work just fine here - as it implicitly does `disableOriginalConstructor` anyway

mimmi20: I updated the tests

Issue #1021, opened 2016-07-14
GetIntent Crawler


clund:
UserAgent: GetIntent Crawler (http://getintent.com/bot.html)

IP: 136.243.75.15

Crawler: true

PR #1026, opened 2016-07-22
add new rule for GetIntent Crawler

mimmi20:
I added a new rule to the GetIntent Crawler and a test. This should solve issue #1021.


asgrim:
Thanks @mimmi20

Issue #1024, opened 2016-07-21
Wrong device detection


podoprygora:
This tablet UA is detecting as mobile:
Mozilla/5.0 (Linux; Android 5.1.1; SAMSUNG SM-P601 Build/LMY47X) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/3.3 Chrome/38.0.2125.102 Safari/537.36

PR #1028, opened 2016-07-22
add Galaxy Note 10.1 to SamsungBrowser

mimmi20:
I added the  Galaxy Note 10.1 to SamsungBrowser, and I added a test. This should solve issue #1024.


asgrim:
Thanks @mimmi20 !

Issue #1031, opened 2016-07-25
Safari 10


rpfennin:
Here are a couple examples of Safari 10 user agents.  Please include in detection.

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12) AppleWebKit/602.1.32 (KHTML, like Gecko) Version/10.0 Safari/602.1.32

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12) AppleWebKit/602.1.38 (KHTML, like Gecko) 
Version/10.0 Safari/602.1.38

PR #1033, opened 2016-07-26
add new macOS 10.12 and safari 10

mimmi20:
I added macOS 10.12 and Safari 10 and two tests. I also added two versions to Apple Mail. Other tests for safari were switched off for the lite mode (this is nessesary because of the changed name MacOSX -> macOS).

This should solve issue #1031.


asgrim:
Thanks @mimmi20

Issue #1032, opened 2016-07-26
Chrome 54 version not being detected


rpfennin:
Example:

Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2794.0 Safari/537.36

asgrim:
Resolved by #1034

PR #1034, opened 2016-07-30
add versions 54 and 55 for Chrome and Chromium and version 32 for Opera

mimmi20:
I added the new versions 54 and 55 for Chrome and Chromium, version 32 for Opera and verions 17 to Edge. I added a test for this.
I removed versions 40 .. 44 of Chrome from the lite file and updated the affected tests.

This should solve issue #1032.


asgrim:
Thanks!

Issue #1045, opened 2016-08-28
UC Browser not reported correctly


mimmi20:
see ua-parser/uap-core#163 for more details.

asgrim:
Fixed by #1050

PR #1050, opened 2016-09-01
add new versions for UC Browser on Windows

mimmi20:
I added two new versions to the UC Browser on Windows. This shouls solve issue #1045.


asgrim:
Thanks @mimmi20

Issue #1046, opened 2016-08-28
add test for Samsung Browser


mimmi20:
Mozilla/5.0 (Linux; Android 5.0.2; SAMSUNG SM-G925F Build/LRX22G) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/4.0 Chrome/44.0.2403.133 Mobile Safari/537.36

asgrim:
Fixed by #1051

PR #1051, opened 2016-09-01
add test for Samsung Browser 4.0

mimmi20:
I added a test for Samsung Browser 4.0. This should solve issue #1046.


asgrim:
Thanks @mimmi20

Issue #1044, opened 2016-08-28
Add few popular mail clients


mimmi20:
see ua-parser/uap-core#159 for more details

asgrim:
Fixed by #1055

PR #1055, opened 2016-09-01
add Mail clients Yahoo Mobile Mail and The Bat

mimmi20:
I added the mail clients Yahoo Mobile Mail and The Bat. And I added some tests. This should solve siiue #1044.


asgrim:
Thanks @mimmi20

Issue #1060, opened 2016-09-08
Missing user agent - Firefox 48 on Windows XP


jsalvata:
`Mozilla/5.0 (Windows NT 5.1; rv:48.0) Gecko/20100101 Firefox/48.0`

PR #1061, opened 2016-09-09
add new test for Firefox 48 on Win XP

mimmi20:
I added a new test for Firefox 48 on Win XP. No other actions were nessesary. This should solve issue #1060.


asgrim:
Thanks @mimmi20!

Issue #1068, opened 2016-09-23
Not detecting or reading Vivaldi correctly


mimmi20:
see serbanghita/Mobile-Detect#584 for more details.

mimmi20:
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.89 Safari/537.36 Vivaldi/1.1.453.47  should be  Vivaldi 1.1.453.47 , but got  Chrome 50.0.2661.89

PR #1089, opened 2016-09-26
add new tests for Vivaldi Browser

mimmi20:
I added two new tests for Vivaldi Browser. This should solve issue #1068.


asgrim:
Thanks for all the PRs @mimmi20 :)

Issue #1112, opened 2016-10-27
Asus memo8, model K011 not detected as tablet


mimmi20:
see serbanghita/Mobile-Detect#592 for more details

PR #1115, opened 2016-10-29
add Asus MeMO Pad 8

mimmi20:
I added the Asus MeMO Pad 8 and a test. This should solve issue #1112.


nvsdesign:
Does this fix automatically apply or do I have to modify the mobile_detect.php file?

asgrim:
@nvsdesign `mobile_detect.php` is not part of this project, so we're unlikely to be able to help... I'd suggest finding the project where that file comes. Thanks!

Issue #1109, opened 2016-10-19
Malformed JSON data 


riyyaz:
Parsing the JSON data from build 6017 produces this error

$cat browscap.json | python -m json.tool
Invalid \escape: line 2879 column 14 (char 559467)

PR #1116, opened 2016-10-29
change encoding when creating json output

mimmi20:
I changed the encoding function when creating the json output. This should solve issue #1109.


src/Browscap/Formatter/JsonFormatter.php
```diff
@@ -128,14 +128,21 @@ private function jsonEncode($val)
...
         }
-        if (is_string($val) || is_numeric($val)) {
-            return '"' . addslashes($val) . '"';
+
+        if (is_string($val)) {
+            return json_encode($val);
+        }
+
+        if (is_numeric($val)) {
+            return '"' . json_encode($val) . '"';
```
asgrim: If it's numeric, why do we need the quotes? Same with `true` and `false`...

Why do we need to granularly write the json vs building an array and just doing `json_encode` on the lot?

Issue #1059, opened 2016-09-02
Missing user agents


patrix1337:
I found missing user agents, most of them are SmartTVs. (Tested with version 6016)
[missingUserAgents_all.txt](https://github.com/browscap/browscap/files/452975/missingUserAgents_all.txt)
[missingUserAgents_filtered.txt](https://github.com/browscap/browscap/files/452976/missingUserAgents_filtered.txt)

PR #1150, opened 2016-12-25
add TV and mobile devices and some App and browsers

mimmi20:
I added some TV devices and mobiles and their related browsers and app.
This should solve issue #1059. Not all useragents of the issue were solved, but the missing uas did not have browser information inside.


asgrim:
Conflicts here @mimmi20 :/

asgrim:
@mimmi20 hrmm - build failed here, looks like changing the device flags to to `unknown` has broken BC somewhere

mimmi20:
@asgrim This is fixed now.

Issue #1155, opened 2017-01-09
iOS 10.x version is "unknown" (safari version is unknown as well)


jaydiablo:
This UA is Safari on iOS 10:

> Mozilla/5.0 (iPhone; CPU iPhone OS 10_1_1 like Mac OS X) AppleWebKit/602.2.14 (KHTML, like Gecko) Version/10.0 Mobile/14B100 Safari/602.1

Browscap 6019 is detecting the iOS version as "unknown" (and the Safari version as "0"):

> browser_name_regex 	/^mozilla\/5\.0 \(iphone.*cpu iphone os .* like mac os x.*\).*applewebkit\/.*\(.*khtml.* like gecko.*\).*version\/.*safari\/.*$/
browser_name_pattern 	mozilla/5.0 (iphone*cpu iphone os * like mac os x*)*applewebkit/*(*khtml* like gecko*)*version/*safari/*
parent 	Mobile Safari Generic
comment 	Mobile Safari Generic
browser 	Safari
browser_type 	Browser
browser_bits 	32
browser_maker 	Apple Inc
browser_modus 	unknown
version 	0.0
majorver 	0
minorver 	0
platform 	iOS
platform_version 	unknown
platform_description 	iPod, iPhone & iPad
platform_bits 	32
platform_maker 	Apple Inc

I ran the user-agent through that comparison tool that @ThaDafinser always links to and do see that some other libraries are detecting both the Safari and iOS version correctly:

http://useragent.mkf.solutions/?userAgent=Mozilla%2F5.0+%28iPhone%3B+CPU+iPhone+OS+10_1_1+like+Mac+OS+X%29+AppleWebKit%2F602.2.14+%28KHTML%2C+like+Gecko%29+Version%2F10.0+Mobile%2F14B100+Safari%2F602.1

We have other user-agents like this, it seems that perhaps the version part of the pattern just needs to be updated to support iOS 10?

Here's a chunk of some we've seen today that are all Safari with an "unknown" iOS version (showing as 10.0, 10.1 or 10.2 in the useragent string):

>Mozilla/5.0 (iPhone; CPU iPhone OS 10_2 like Mac OS X) AppleWebKit/602.3.12 (KHTML, like Gecko) Version/10.0 Mobile/14C92 Safari/602.1

[...]

PR #1174, opened 2017-01-20
add Safari Mobile 10, add iOS 10.x

mimmi20:
I added iOS 10.1 and 10.2 and Safari Mobile 10.x and a test. this should solve issue #1155.


mimmi20:
@asgrim Could you check and merge this PR? This fixes an error which is actually in the mester.

Issue #1223, opened 2017-02-20
UC Browser detected as Android


jaydiablo:
Have a handful of useragent strings that should probably be detected as "UC Browser" but instead are being detected as "Android".  This is against the Full file, version 6021.  I don't believe any of the commits since 6021 was released have addressed UC Browser agents at all.

These are agents we've seen in our logs, and we believe them to be legit UC Browser agents due to the proxy they connected from.

>Mozilla/5.0 (Linux; U; Android 4.4.2; en-US; ASUS_Z007 Build/KVT49L) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 UCBrowser/11.2.0.915 U3/0.8.0 Mobile Safari/534.30
Mozilla/5.0 (Linux; U; Android 5.1.1; en-US; N9132 Build/LMY47V) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 UCBrowser/11.2.0.915 U3/0.8.0 Mobile Safari/534.30
Mozilla/5.0 (Linux; U; Android 5.1.1; en-US; SM-J500F Build/LMY48B) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 UCBrowser/11.1.0.882 U3/0.8.0 Mobile Safari/534.30
Mozilla/5.0 (Linux; U; Android 5.1.1; en-US; SM-J200G Build/LMY47X) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 UCBrowser/11.2.0.915 U3/0.8.0 Mobile Safari/534.30
Mozilla/5.0 (Linux; U; Android 5.0.2; en-US; SM-G530H Build/LRX22G) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 UCBrowser/11.1.0.882 U3/0.8.0 Mobile Safari/534.30
Mozilla/5.0 (Linux; U; Android 4.4.2; en-US; ASUS_T00I Build/KVT49L) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 UCBrowser/11.2.0.915 U3/0.8.0 Mobile Safari/534.30
Mozilla/5.0 (Linux; U; Android 5.1; en-US; Aqua Trend Build/LMY47D) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 UCBrowser/11.1.5.890 U3/0.8.0 Mobile Safari/534.30
[...]

PR #1241, opened 2017-02-22
Adding UC Browser versions and new devices for #1223

jaydiablo:
This adds UC Browser versions 11.1 and 11.2 (which all of the
useragents in #1223 are). I also added a lot of new devices for these
strings (a couple of them had a device identifier that wasn’t very
specific, so I left those ones as “general Mobile Phone”).  Test added
for each agent in that ticket, all tests pass.


resources/user-agents/browsers/uc-browser/uc-web-11.json
```diff
@@ -1,6 +1,6 @@
 {
   "division": "UC Browser #MAJORVER#.#MINORVER#",
-  "versions": ["11.0"],
+  "versions": ["11.0","11.1","11.2"],
```
mimmi20: Please change the order for the versions. Please start with the newest version.

PR #1248, opened 2017-02-24
Converting UC Browser 3 JSON file to use "devices" array where possible

jaydiablo:
Combined like patterns/engines/platforms into single entries with a
devices array where possible.

Also, there was a duplicate pattern here, so I removed that one.

>UCWEB\/*(#PLATFORM#SAMSUNG; GT-I8350*) U2\/*
UCBrowser\/#MAJORVER#.#MINORVER#* U2\/*

### DIFF
--- a/resources/user-agents/browsers/uc-browser/uc-web-3.json
+++ b/resources/user-agents/browsers/uc-browser/uc-web-3.json
@@ -65,9 +65,19 @@
           ]
         },
         {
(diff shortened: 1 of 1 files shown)


resources/user-agents/browsers/uc-browser/uc-web-3.json
```diff
@@ -255,15 +177,6 @@
           ]
         },
         {
-          "match": "UCWEB\/*(#PLATFORM#SAMSUNG; GT-I8350*) U2\/* UCBrowser\/#MAJORVER#.#MINORVER#* U2\/*",
```
jaydiablo: This is the identical pattern (to the one above it). I triple checked the pattern, engines, device, platforms.  Tests still pass with it removed.

Issue #1244, opened 2017-02-23
New app UA


zahra-travail:
I've got an unknown (under Browscap):

`"AppName"/3.7.2 (Linux; U; Android 4.1.1; en-ca; Arc Build/JRO03L) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 Safari/534.30`

WhichBrowser and NeutrinoApiCom seem to have the most complete information on this UA, but I couldn't suss out the regex for this one.

http://useragent.mkf.solutions/?userAgent=RedFlagDealsApp%2F3.7.2+%28Linux%3B+U%3B+Android+4.1.1%3B+en-ca%3B+Arc+Build%2FJRO03L%29+AppleWebKit%2F534.30+%28KHTML%2C+like+Gecko%29+Version%2F4.0+Safari%2F534.30

PR #1259, opened 2017-02-26
add RedFlagDeals Mobile App

mimmi20:
I added the Kobo Arc and the RedFlagDeals Mobile App and a test. This should solve sissue #1244.


resources/devices.json
```diff
@@ -30759,7 +30759,7 @@
       }
     },
     "Xiaomi Mi Pad": {
-      "standard": false,
+      "standard": true,
```
asgrim: All devices that are changed to `standard` here - should they really be? I've only heard of Google Pixel, but that's not to say they shouldn't be in `standard` - just checking they are "useful" enough in general?

mimmi20: Tablets need to be in standard mode to have the `isTablet` Property.

Issue #1219, opened 2017-02-15
Potential pattern cleanup


jaydiablo:
More discussion in https://github.com/browscap/browscap-php/issues/179.

These are useragents that `browscap-php` and PHP's `get_browser` function interpret differently (currently).  @mimmi20 suggested that the patterns could use some cleanup.

Here's the first list of 18 from this comment:  https://github.com/browscap/browscap-php/issues/179#issuecomment-279798423

```
Mozilla/4.0 (compatible; MSIE 4.01; Mac_PowerPC)
Mozilla/5.0 (Linux; U; Android 2.3.7; de-de; HTC DESIRE HD Build/GRI40; SUNDAWG CM7) AppleWebKit/533.1 (KHTML, like Gecko) Version/4.0 Mobile Safari/533.1
Mozilla/5.0 (Linux; U; Android 2.3.3; en-cn; HTCA510e/1.0 Android/2.4 release/02.25.2011 Browser/WAP 2.0 Profile/MIDP-2.0 Configuration/CLDC-1.1 Build/GRI40) AppleWebKit/533.1 (KHTML, like Gecko) Version/4.0 Mobile Safari/533.1
Mozilla/5.0 (Linux; U; Android 2.2.2; en-us; HUAWEI T8600 Build/FRF91) AppleWebKit/533.1 (KHTML, like Gecko) FlyFlow/1.0 Version/4.0 Mobile Safari/533.1
Mozilla/5.0 (Linux; Android 3.1; pt-PT; MZ606 Build/UMWB8E) AppleWebKit/535.19 (KHTML, like Gecko) Chrome/18.0.1025.166 Mobile Safari/535.19
Mozilla/5.0 (Linux; Android 3.2; en-US; SHW-M305W Build/P2FHU4) AppleWebKit/535.19 (KHTML, like Gecko) Chrome/18.0.1025.166 Mobile Safari/535.19
NetFront/3.5.1(BREW 3.1.5; U; en-us; SAMSUNG; NetFront/3.1.5/WAP) PLS-M350 MMP/2.0 Profile/MIDP-2.1 Configuration/CLDC-1.1
NetFront/3.5.1(BREW 3.1.5; U; en-us; SAMSUNG; NetFront/3.1.5/WAP) Sprint M350 MMP/2.0 Profile/MIDP-2.1 Configuration/CLDC-1.1
[...]

PR #1266, opened 2017-03-05
Reformat webkit token

mimmi20:
There are different writings of the `Webkit` token and of the `KHTML` token in the resource files. 

This differences are with/without leading/trailing star `*`, with/wihout comma between `KHTML` and `like Gecko`, missing space between comma and `like` or between `like` and `Gecko`.

This PR intends to remove these differences.

This PR is required to solve issze #1219.


asgrim:
Conflicts here to be resolved please @mimmi20

mimmi20:
@asgrim The conflicts are solved.

Issue #1238, opened 2017-02-22
Recent Versions of iOS detected as Mobile regardless of Device


zahra-travail:
Using https://browscap.org/ua-lookup

UA such as the following are detected as a "General Mobile Device" instead of iPad/iPhone/iPod as appropriate.

`Mozilla/5.0 (iPad; U; CPU OS 10_2 like Mac OS X; en-us; iPad4,7) AppleWebKit/534.46 (KHTML, like Gecko) UCBrowser/2.4.0.367 U3/1 Safari/7543.48.3`

Correct parsing should be similar (change as appropriate for iPod and iPhone) to
`Mozilla/5.0 (iPad; CPU OS 10_0 like Mac OS X) AppleWebKit/602.1.43 (KHTML, like Gecko) Version/10.0 Mobile/14A5322e Safari/602.1`

PR #1269, opened 2017-03-07
add new test for UCBrowser on iPad

mimmi20:
I added new iOS versions to the UCBrowser for the iPad. This should solve issue #1238.


asgrim:
Fixes #1238

PR #1274, opened 2017-03-09
Splitting devices.json file up

jaydiablo:
I put together a quick script that breaks up the devices.json file.

First by device type, then by brand name. I didn’t do any grouping of
“other” agents or anything like that, so it does end up making a lot of
single device files.

Let me know if you’d like to see what this looks like with different
groupings.


mimmi20:
Having a lot of single device files should not be an issue.

Issue #1135, opened 2016-12-06
Pre-generate the files when running Travis builds for speed


asgrim:
If we pre-generate the files for the build instead of building them whilst running PHPUnit, we can see massive performance increases.

Basically:

 * Use `php -n` to disable any ini files (thus, xdebug does not get loaded) and generate the files
 * THEN run `vendor/bin/phpunit` (but don't generate the files if they already exist).

If the files don't exist, they should still be generated so the existing behaviour is maintained though!

mimmi20:
The second `phpunit` call could be run without xdebug because we call it with the `--no-coverage` option.

asgrim:
Yep, but it's the generation of the files (done in the first `phpunit` run) that takes the time (~10m with xdebug, ~2.5m without)

mimmi20:
@asgrim Is this issue still open? It should be fixed with PR #1222.

PR #1279, opened 2017-03-10
Add pattern coverage with codecov.io

### LINKED ISSUE #1253: Pattern test coverage
So as I'm condensing some of these patterns in various files (no particular method to the madness at the moment), I was a bit concerned that I may be modifying patterns incorrectly (even though I'm being careful not to). I was (incorrectly) under the assumption that if patterns are added to the resources, they've had a test added as well, but as I realized in #1251 this isn't always the case.

I was curious how many patterns exist in the INI file that aren't touched by a test from the test suite.

I extracted the patterns from the INI, but turned all digits into hash signs (#) in order to make the list a bit more manageable/realistic (I don't think it's terribly important that every possible version of every possible pattern is touched by a test, that would be a bit insane).  I then ran the test suite, and extracted all patterns that were used against the user agents in the suite (again, replacing all digits with "#").

I get 13,811 unique patterns (with digits removed) from the INI file (this is taken from the update step of browscap-php, so it doesn't include the "group" patterns).

[available_patterns.txt](https://github.com/browscap/browscap/files/800795/available_patterns.txt)

And from the test suite I get 3,467 patterns used.

[tested_patterns.txt](https://github.com/browscap/browscap/files/800796/tested_patterns.txt)

If I'm doing this right, that leaves a lot of untested patterns! There are of course a lot of small variations in patterns aside from the version numbers (i.e. minor platform/device differences). And as I understand it, a lot of the files have a sort of "catch-all" pattern(s) that are used to at least identify the browser with basic properties based on the UA containing a certain structure/keywords for that browser.  These probably don't typically have a test UA to go with them.

Anyhow, I was curious what it would look like if I ran browscap-php against a large list of user-agents, and only dumped the patterns/uas that browscap doesn't cover during its own tests.

The list I used comes from various other parsers' test suites (about 47k agents). This dumped 1,761 unique, uncovered patterns (with digits replaced with "#").

[all-suites-uncovered-patterns.txt](https://github.com/browscap/browscap/files/800808/all-suites-uncovered-patterns.txt)

And I've also extracted just one useragent string for each pattern and assembled those into a text file:

[useragents-from-uncovered-patterns.txt](https://github.com/browscap/browscap/files/800810/useragents-from-uncovered-patterns.txt)

If you're at all curious to see **all** of the useragents collected during this process (that were parsed using an uncovered pattern), here's that list, grouped by each pattern:

[all-useragents-from-uncovered-patterns.txt](https://github.com/browscap/browscap/files/800814/all-useragents-from-uncovered-patterns.txt)

Is there much interest in adding some of these useragents to the test-suite to increase the coverage of the patterns?  Certainly a tedious task.

Are there any built-in tools to analyze this better than the approach that I took?

I could do this against a more "real world" list of agents if that seems like a better approach.

jaydiablo:
This fixes #1253, a lot more discussion there.

This PR adds pattern id generation and collection to the full user agent test suite so that a coverage report can be built against the resource JSON files, showing which patterns (and their different expansions) have tests that touch them.  I feel this is important when doing major cleanup/refactoring of the JSON files, as it gives more confidence that the tests will catch any mistakes.

Currently, the coverage report is sent to codecov.io (my branch's report here: https://codecov.io/gh/jaydiablo/browscap/tree/05c9536d2de6adb7275030bb90182493e46da7fc), which I feel has a better UI for quickly scanning which directories/files are covered or not.  I do have a branch of this branch that adds coveralls support (https://github.com/jaydiablo/browscap/tree/pattern-coverage-coveralls), which could still be added or swapped out if desired.  Codecov.io does have a github integration that should probably be added to the browscap/browscap repo after this is merged in order to enable some of the automatic PR checks and bot commenting mentioned in #1253.  Not a requirement to publish and view the coverage reports on codecov.io though.

A local coverage report can also be built using Istanbul. First, istanbul has to be installed: `npm install -g istanbul`. Then, after the useragent test suite is run (and a coverage.json file is created) you can run `istanbul report` to build a local HTML report for viewing (HTML report gets written to `./coverage/lcov-report/index.html`.

When pattern id generation is enabled (triggered by the full useragent suite, or by a new `--coverage` command line option to the `build` command) pattern ids will be collected and inserted into the resulting INI file (as the "PatternId" property for each match).  The pattern ids are in this format: `<resource_file_name>::u<position_in_userAgents_array>::c<position_in_children_array>::d<device_match_name_if_present>::p<platform_name_if_present>` This gives us enough information to be able to determine the position of the pattern that was touched by a test when the JSON files are later processed. 

The full useragent test suite collects any PatternIds that are encountered by the tests, and then passes those to the JSON coverage processor which are used to determine which lines/branches of the JSON file were touched.

I couldn't think of a really good way to disable the pattern id generation/collection and JSON coverage building, so currently it happens whenever the full useragent test is run.  This does add some time to the test run (mainly when generating the coverage.json file).  Initially I thought this would only be done when coverage is enabled (calling the `xdebug_code_coverage_started` function and checking if the result is `true`), but later realized that coverage is disabled for the useragent tests in the travis builds.  We could explore using an environment variable or something like that to enable/disable the coverage report generation if that makes sense (I believe travis can set a variable, or call a script that sets the variable before tests are run).  I don't think it's possible to define our own CLI option to the `phpunit` command, as PHPUnit seems to validate the CLI options and will bail if it doesn't recognize the option.

I've tested and confirmed that the PatternIds are not inserted into the INI, JSON, CSV, or XML files when using the build command without the `--coverage` option.  When using that option, only the INI files get the PatternIds.  I added that option to the build command in case it was needed for something like #1135.

As for the coverage report itself, since we're covering JSON data files rather than executable Javascript code, I had to translate certain structures in the browscap resource files to an equivalent in language syntax. Each "match" section of a child is considered a function, and then each platform (from the `platforms` array) and each device (only from the `devices` hash) are considered a branch (specifically a switch statement). Codecov.io doesn't really break down the statement/branch/function coverage in the UI, but the Istanbul report does:

screen shot 2017-03-10 at 11 38 07 am

I've tried to make sure the coverage.json file is as similar to an actual coverage report from javascript code as possible.  Codecov.io still doesn't render the report quite as I'm expecting (mentioned in #1253), and they haven't responded to my inquiry about that yet. If they do come back with changes I should make to enable the better report, I'll make the necessary changes and submit an additional PR (if this one has been merged already).

I've also added tests to cover some of the changes/additions that I've made to the PHP src files.

Let me know if there are any questions/concerns about all of this, I'm pretty pleased with the way the coverage report has turned out, and have already found it useful when doing some of the device expansion PRs I've done.


codecov.yml
```diff
@@ -0,0 +1,11 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+comment:
+  require_changes: yes
```
jaydiablo: This will make the comment bot less noisy I've noticed, hard to say for sure until it's operating on master branch which has a base report to compare to. May need to be changed in the future.

Docs here: https://docs.codecov.io/docs/pull-request-comments

resources/core/default-browser.json
```diff
@@ -53,7 +53,8 @@
         "RenderingEngine_Name": "unknown",
         "RenderingEngine_Version": "unknown",
         "RenderingEngine_Description": "unknown",
-        "RenderingEngine_Maker": "unknown"
+        "RenderingEngine_Maker": "unknown",
+        "PatternId": "resources/core/default-browser.json::u0"
```
jaydiablo: I needed to add a hard-coded pattern id to these two files in order for the pattern ids to appear in the other patterns (the code checks that default properties specifically has the key). The changes I've made to the BuildHelper class will remove these if PatternId collection isn't enabled.

src/Browscap/Data/DataCollection.php
```diff
@@ -467,7 +467,8 @@ public function addSourceFile($src)
             $userAgents,
             (boolean) $divisionData['lite'],
             (boolean) $divisionData['standard'],
-            $versions
+            $versions,
+            substr($src, strpos($src, 'resources/'))
```
jaydiablo: Normalizing the filename a bit so that everything before "resources/" is removed.

tests/fixtures/issues/issue-1253.php
```diff
@@ -0,0 +1,216 @@
+<?php
```
jaydiablo: Oh, forgot to mention this file.  This is just adding a few Kindle useragents to the test suite to increase coverage of the kindle.json file. I mainly did this as a test of how codecov.io showed a change in coverage from one commit to the next. I felt it was fine to leave in, but can remove and re-submit as a separate PR if desired.

jaydiablo:
FYI, this PR made a coverage report on codecov.io: https://codecov.io/gh/browscap/browscap/tree/05c9536d2de6adb7275030bb90182493e46da7fc

mimmi20:
@jaydiablo Do you know why the codecov checks failed?

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+     */
+    public function write($fileName)
+    {
+        file_put_contents(
+            $fileName,
+            // str_replace here is to convert empty arrays into empty JS objects, which is expected by
+            // codecov.io. Owner of the service said he was going to patch it, haven't tested since
+            // Note: Can't use JSON_FORCE_OBJECT here as we **do** want arrays for the 'b' structure
+            // which FORECE_OBJECT turns into objects, breaking at least the Istanbul coverage reporter
+            str_replace('[]', '{}', json_encode($this->coverage, JSON_UNESCAPED_SLASHES))
```
mimmi20: could we use the option JSON_FORCE_OBJECT to ommit the string replace?

jaydiablo: As mentioned in the comment there, no. The reason being, there are array structures in the coverage.json file that need to remain as arrays.  In the "b" structure (which is where the coverage data for different branches are stored) each entry is an array showing which branches of the branch statement are covered.  Which basically looks like this:

```json
"b": {
  "1": [0,1,2],
  "2": [0,0]
}
```

If I use the JSON_FORCE_OBJECT option, that turns into something like this:

```json
"b": {
  "1": {"0": 0, "1": 1, "2": 2},
  "2": {"0": 0, "1": 0}
}
```

Which makes the Istanbul reporter throw an error:

>~/.npm_packages/lib/node_modules/istanbul/lib/report/lcovonly.js:79
>            branchArray.forEach(function (b) {
>                        ^
>
>TypeError: branchArray.forEach is not a function

`branchArray` in this case is those sections in the "b" hash that should be arrays.

jaydiablo:
@mimmi20 regarding the codecov.io check failing there, I believe it's because it doesn't have a coverage report to compare against, because no coverage report was ever created for the master branch (that this PR is going against).

In the future, once/if this is merged, codecov.io should generate a coverage report for browscap/browscap:master (on the merge commit), and then use that as a comparison moving forward.  I believe then the checks would show how much the coverage changes for any new PRs, as seen in their docs:

commit status example

https://docs.codecov.io/docs/commit-status

Unrelated, but I just noticed this. Apparently there's a browser extension that you can install that will overlay coverage reports directly in github source browsing: https://docs.codecov.io/docs/browser-extension

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+    /**
+     * @var string
+     */
+    private $resourceDir;
+
+    /**
+     * These are available in the JSON Parser part of the JSON Lint package we're using
+     * to Lexically analyze the JSON file, but they're a private property so can't re-use them directly
+     *
+     * @var array
```
asgrim: `@var int[]`

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+        ',' => 22,
+        '[' => 23,
+        ']' => 24,
+        'JSONElementList' => 25,
+        '$accept' => 0,
+        '$end' => 1,
+    ];
+
+    /**
+     * @var array
```
asgrim: Can we be more specific about types please?

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+    {
+        $this->resourceDir = $resourceDir;
+
+        $this->symbolLookup = array_flip($this->symbols);
+    }
+
+    /**
+     * Process the directory of JSON files using the collected pattern ids
+     *
+     * @param array $coveredIds
```
asgrim: I'm assuming `@param int[]` ?

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+     * @var int
+     */
+    private $funcCount = 0;
+
+    /**
+     * Create a new Coverage Processor for the specified directory
+     *
+     * @param string $resourceDir
+     */
+    public function __construct($resourceDir)
```
asgrim: `string $resourceDir` (PHP 7 is minimum)

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+            );
+        }
+    }
+
+    /**
+     * Write the coverage data in JSON format to specified filename
+     *
+     * @param string $fileName
+     */
+    public function write($fileName)
```
asgrim: `string $fileName`

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+                file_get_contents($file->getPathname()),
+                $this->coveredIds[$patternFileName]
+            );
+        }
+    }
+
+    /**
+     * Write the coverage data in JSON format to specified filename
+     *
+     * @param string $fileName
```
asgrim: Missing `@return void` (should add this to make migration to PHP 7.1 easier eventually, or alternatively, I am more than happy to bump minimum to 7.1, but not in this PR - in which case ` : void` return type declaration to be added)

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+            str_replace('[]', '{}', json_encode($this->coverage, JSON_UNESCAPED_SLASHES))
+        );
+    }
+
+    /**
+     * Stores passed in pattern ids, grouping them by file first
+     *
+     * @param array $coveredIds
+     */
+    public function setCoveredPatternIds(array $coveredIds)
```
asgrim: `: void`

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+    {
+        $this->coveredIds = $this->groupIdsByFile($coveredIds);
+    }
+
+    /**
+     * Returns the grouped pattern ids previously set
+     *
+     * @return array
+     */
+    public function getCoveredPatternIds()
```
asgrim: `: array`

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+
+    /**
+     * Process an individual file for coverage data using covered ids
+     *
+     * @param  string $file
+     * @param  string $contents
+     * @param  array  $coveredIds
+     * @return array
+     */
+    public function processFile($file, $contents, array $coveredIds)
```
asgrim: `(string $file string $contents, array $coveredIds) : array`

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+        return $this->coveredIds;
+    }
+
+    /**
+     * Process an individual file for coverage data using covered ids
+     *
+     * @param  string $file
+     * @param  string $contents
+     * @param  array  $coveredIds
+     * @return array
```
asgrim: `array` of what?

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+    {
+        return $this->coveredIds;
+    }
+
+    /**
+     * Process an individual file for coverage data using covered ids
+     *
+     * @param  string $file
+     * @param  string $contents
+     * @param  array  $coveredIds
```
asgrim: `array` of what?

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+        if (strlen($p) === 0) {
+            $p = '.*?';
+        }
+        if (strlen($d) === 0) {
+            $d = '.*?';
+        }
+
+        $regex = sprintf('/^u%d::c%d::d%s::p%s$/', $u, $c, $d, $p);
+
+        //print $regex . PHP_EOL;
```
asgrim: Remove commented lines of code

src/Browscap/Coverage/Processor.php
asgrim: This method is really long, we we should really break this down into meaningful `private` methods that describe clearly what is going on (i.e. describe intent)

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+            $line    = $lines[$lexer->yylineno];
+
+            switch ($state) {
+                case '':
+                    if ($type === '{') {
+                        $state = 'inDivision';
+                        continue;
+                    }
+                    break;
+                case 'inDivision':
```
asgrim: Could we define constants for these to assist refactoring if necessary please?

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+                    if ($type === 'STRING' && $content === 'userAgents') {
```
asgrim: More constants to be defined for these (definitely `$type`, possibly `$content` depending what it's used for)

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+        return $coverage;
+    }
+
+    /**
+     * Groups pattern ids by their filename prefix
+     *
+     * @param  array $ids
+     * @return array
+     */
+    private function groupIdsByFile(array $ids)
```
asgrim: More type declarations / clarification on `array` contents please

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+    }
+
+    /**
+     * Counts number of times given pattern is covered by test patterns
+     *
+     * @param  string $id
+     * @param  array  $covered
+     * @return int
+     */
+    private function getCoverageCount($id, array $covered)
```
asgrim: `int` and what is in this `array`?

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+        $count = 0;
+
+        if (strlen($p) === 0) {
+            $p = '.*?';
+        }
+        if (strlen($d) === 0) {
+            $d = '.*?';
+        }
+
+        $regex = sprintf('/^u%d::c%d::d%s::p%s$/', $u, $c, $d, $p);
```
asgrim: need to `preg_quote` these values

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+     * Counts number of times given pattern is covered by test patterns
+     *
+     * @param  string $id
+     * @param  array  $covered
+     * @return int
+     */
+    private function getCoverageCount($id, array $covered)
+    {
+        $id                  = str_replace('\/', '/', $id);
+        list($u, $c, $d, $p) = explode('::', $id);
```
asgrim: What are `$u`, `$c`, `$d` and `$p`? They tell me nothing about what they're doing

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+
+use Seld\JsonLint\Lexer;
+
+/**
+ * Class Processor
+ *
+ * @category   Browscap
+ * @author     Jay Klehr <jay.klehr@gmail.com>
+ */
+class Processor
```
asgrim: This should be `final` and implement an interface (to be created)

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
+<?php
```
asgrim: `declare(strict_types=1);` please

composer.json
```diff
@@ -40,7 +40,8 @@
     "browscap/browscap-php": "dev-master",
     "mikey179/vfsStream": "^1.6",
     "squizlabs/php_codesniffer": "^2.8",
-    "friendsofphp/php-cs-fixer": "^1.11"
+    "friendsofphp/php-cs-fixer": "^1.11",
+    "seld/jsonlint": "^1.5"
```
asgrim: I've realised `composer.lock` is gitignored - please can we remove that from gitignore and commit along with the repo. This is a project, not a library so we should have locked down deps to avoid issues.

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+            'fnMap' => [],
+            'branchMap' => [],
+            's' => [],
+            'b' => [],
+            'f' => [],
+        ];
+
+        $lines = explode("\n", $contents);
+
+        $u = null;
```
asgrim: These variable names are meaningless. Please refactor to show what they mean

src/Browscap/Coverage/Processor.php
```diff
@@ -0,0 +1,571 @@
...
+     * @return array
+     */
+    public function processFile($file, $contents, array $coveredIds)
+    {
+        $coverage = [
+            'path' => $file,
+            'statementMap' => [],
+            'fnMap' => [],
+            'branchMap' => [],
+            's' => [],
```
asgrim: Can we rename these keys to something more meaningful?

tests/BrowscapTest/Coverage/ProcessorTest.php
```diff
@@ -0,0 +1,182 @@
...
+        self::assertSame($expected['functionCount'], count($coverage['f']));
+    }
+
+    /**
+     * Data provider for the testCoverage test
+     */
+    public function coverageProvider()
+    {
+        return [
+            // No coverage
```
asgrim: Rather than commenting here, you can turn these into useful debugging info for phpunit by making them the array keys (then phpunit uses them to identify which iteration of the provider we're on with meaningful name)

tests/BrowscapTest/Coverage/ProcessorTest.php
```diff
@@ -0,0 +1,182 @@
...
+            ['test2.json', ['statementCount' => 15, 'branchCount' => 3, 'functionCount' => 3]],
+        ];
+    }
+
+    /**
+     * This test verifies that the different structures were extracted from the test JSON files
+     *
+     * @dataProvider jsonStructureProvider
+     */
+    public function testJsonStructure($fileName, $expected)
```
asgrim: Types

tests/BrowscapTest/Coverage/ProcessorTest.php
```diff
@@ -0,0 +1,182 @@
...
+            ],
+        ];
+    }
+
+    /**
+     * Tests that the amount of covered statements/branches/functions matches expected
+     *
+     * @dataProvider coverageProvider
+     */
+    public function testCoverage($fileName, $coveredIds, $expected)
```
asgrim: Types

src/Browscap/Data/Division.php
```diff
@@ -61,15 +66,17 @@ class Division
      * @param bool   $lite
      * @param bool   $standard
      * @param array  $versions
+     * @param string $fileName
      */
-    public function __construct($name, $sortIndex, array $userAgents, $lite, $standard = true, array $versions = [])
+    public function __construct($name, $sortIndex, array $userAgents, $lite, $standard = true, array $versions = [], $fileName = null)
```
asgrim: Types to be added pleas

src/Browscap/Data/Division.php
```diff
@@ -119,4 +126,12 @@ public function getVersions()
     {
         return $this->versions;
     }
+
+    /**
+     * @return string
+     */
+    public function getFileName()
```
asgrim: Return type to be added please

src/Browscap/Data/Expander.php
```diff
@@ -43,6 +43,11 @@ class Expander
     private $logger = null;
 
     /**
+     * @var array
```
asgrim: Array of...? `@var int[]` perhaps?

src/Browscap/Data/Expander.php
```diff
@@ -368,6 +407,25 @@ private function parseChildren($ua, array $uaDataChild, $lite = true, $standard
     }
 
     /**
+     * Builds and returns the string pattern id from the array components
+     *
+     * @return string
+     */
+    private function getPatternId()
```
asgrim: `: string`

src/Browscap/Data/Expander.php
```diff
@@ -368,6 +407,25 @@ private function parseChildren($ua, array $uaDataChild, $lite = true, $standard
...
+    {
+        $id = sprintf(
```
asgrim: No need to declare `$id`, just return the value from `sprintf` directly

src/Browscap/Generator/AbstractBuildGenerator.php
```diff
@@ -87,6 +92,20 @@ public function getLogger()
     }
 
     /**
+     * Sets the flag to collect pattern ids during this build
+     *
+     * @param bool $value
+     *
+     * @return \Browscap\Generator\AbstractBuildGenerator
+     */
+    public function setCollectPatternIds($value)
```
asgrim: `bool $value` and `: self`

asgrim: Though, I disagree with fluent method interfaces, it **is** consistent with existing implementation.

src/Browscap/Generator/Helper/BuildHelper.php
```diff
@@ -47,7 +47,8 @@ public static function run(
         $resourceFolder,
         LoggerInterface $logger,
         WriterCollection $writerCollection,
-        CollectionCreator $collectionCreator
+        CollectionCreator $collectionCreator,
+        $collectPatternIds = false
```
asgrim: `bool $collect...`

src/Browscap/Generator/Helper/BuildHelper.php
```diff
@@ -47,7 +47,8 @@ public static function run(
...
     ) {
```
asgrim: Add a return type declaration

tests/BrowscapTest/Coverage/ProcessorTest.php
```diff
@@ -0,0 +1,182 @@
+<?php
```
asgrim: `declare(strict_types=1);`

tests/BrowscapTest/Coverage/ProcessorTest.php
```diff
@@ -0,0 +1,182 @@
...
+
+use Browscap\Coverage\Processor;
+
+/**
+ * Class ExpanderTest
+ *
+ * @category   BrowscapTest
+ * @author     Jay Klehr <jay.klehr@gmail.com>
+ */
+class ProcessorTest extends \PHPUnit\Framework\TestCase
```
asgrim: `final`

src/Browscap/Coverage/Processor.php
jaydiablo: These keys are what will be in the coverage.json file, which is what Istanbul/codecov.io are expecting.  If we name them something else here, we'd have to add a conversion step to convert them to the keys that the reporters are expecting.

src/Browscap/Coverage/Processor.php
jaydiablo: I used $u, $p, $d and $c to match the letter prefix used in the PatternId, since they're used to represent the data of that spot in the id.  Still like me to change?

Most likely I would change to something like this:

$u -> $userAgentPosition
$c -> $childNumber
$d -> $deviceMatch
$p -> $platformName

src/Browscap/Coverage/Processor.php
jaydiablo: $d and $p are quoted above, but I'll do the same the $u and $c.

tests/BrowscapTest/Coverage/ProcessorTest.php
jaydiablo: None of the other BrowscapTest classes are final, still want me to change this?

src/Browscap/Data/Expander.php
jaydiablo: This one is mixed, some are int, some are string.

src/Browscap/Coverage/Processor.php
jaydiablo: This one ends up being the big array that is json_encoded and written to the file, it ends up being a multi-dimensional array (with the keys seen below; `fnMap`, `statementMap`, etc...).

src/Browscap/Coverage/Processor.php
jaydiablo: again, multi-dimensional array (that can't be simplified to `string[]`, `int[]`, etc... right?)

src/Browscap/Coverage/Processor.php
asgrim: Ok - can we document at least the top level keys and their meaning please?

src/Browscap/Coverage/Processor.php
asgrim: If the values are all the same, you can; e.g. if you have array structure:

```php
[
  [1, 2, 3],
  [4, 5, 6],
];
```
The hint can be `int[][]`

If the structure can be defined clearly, wondering if it makes sense to wrap it in a DTO; would hinting and strictness benefit from that?

src/Browscap/Coverage/Processor.php
asgrim: OK; please add a comment to that effect (describing what they mean also)

src/Browscap/Coverage/Processor.php
asgrim: Yes please, much clearer what they mean now :)

tests/BrowscapTest/Coverage/ProcessorTest.php
asgrim: Yes please; when I get some time I'd like to fix all sorts of little things like this out :)

src/Browscap/Data/Expander.php
asgrim: OK no problem; same question, would it benefit a DTO to be used here to enforce structure?

src/Browscap/Coverage/Processor.php
jaydiablo: So I added `@return void` to a lot of the changes in this PR for methods that don't return anything, and then ran `php-cs-fixer` on everything, which then removed them all due to this rule: `phpdoc_no_empty_return`, defined in the fixer config.

I'll remove that rule in php-cs-fixer config if there are no objections.

composer.json
jaydiablo: composer.lock removed from .gitignore and composer.lock added to latest commit.

src/Browscap/Coverage/Processor.php
jaydiablo: added

src/Browscap/Coverage/Processor.php
jaydiablo: `@return void`s added where applicable (and php-cs-fixer rule removed so they don't get removed when fixer is run).

src/Browscap/Coverage/Processor.php
jaydiablo: quoted

src/Browscap/Data/Expander.php
jaydiablo: Could turn this into an object I suppose, and type-hint with that.  My only concern would be the added overhead to do that compared to a simple associative array here, as this id is built for every pattern that is encountered in the useragent resource files.

tests/BrowscapTest/Coverage/ProcessorTest.php
jaydiablo: done

src/Browscap/Coverage/Processor.php
asgrim: Yes, that should've been gone :) nice one :+1:

src/Browscap/Data/Expander.php
asgrim: data structures are hard :) let's leave it as is then, is it possible to comment on what the structure should look like at least?

jaydiablo: I could set it here to the empty state that is seen below (line 102). I didn't do that initially as I thought it was unnecessary duplication.

src/Browscap/Data/Expander.php
```diff
@@ -79,7 +88,7 @@ public function getVersionParts($version)
      * @throws \UnexpectedValueException
      * @return array
      */
-    public function expand(Division $division, $divisionName)
+    public function expand(Division $division, $divisionName) : array
```
asgrim: `string $divisionName` please

tests/BrowscapTest/Coverage/ProcessorTest.php
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /**
+     * This test verifies that the different structures were extracted from the test JSON files
+     *
+     * @dataProvider jsonStructureProvider
+     *
+     * @return void
+     */
+    public function testJsonStructure($fileName, $expected)
```
asgrim: types here please

tests/BrowscapTest/Generator/Helper/BuildHelperTest.php
```diff
@@ -69,6 +69,10 @@ public function testRun()
             ->will(self::returnSelf());
         $writerCollection->expects(self::any())
             ->method('renderSectionBody')
+            ->with(self::callback(function ($props) {
```
asgrim: `array` type please

jaydiablo:
I'm not really sure why the travis build won't trigger. It's weird that it doesn't even show the details link. The first time I noticed it, I pushed another change, and the build kicked off fine.  But past two edits haven't.

jaydiablo:
I wonder if the lack of travis build has to do with me having an open PR in my own fork (I was testing the codecov.io integration there): https://github.com/jaydiablo/browscap/pull/2

jaydiablo:
I closed that PR on my fork before pushing that commit, looks like Travis ran.  Maybe that was the issue. *shrug*

asgrim:
@jaydiablo really big thanks for this, nice work getting coverage in :)

jaydiablo:
Thanks for reviewing/merging, looking forward to using this.

I spoke with Steve @ codecov.io yesterday, he's still looking into the rendering issues I mentioned, but sounds confident he can fix it. :)

PR #1283, opened 2017-03-14
Fixing and preventing files from getting multiple platforms/devices without the #PLATFORM# or #DEVICE# token in the match

jaydiablo:
I noticed this earlier with the WhatsApp file and fixed it then. Figured I’d do a quick scan to see if it happens anywhere else and found a few more instances.

This PR fixes those cases (one of them I added the #PLATFORM# token, the other one I removed the second platform since it didn’t really make sense there).  It also adds some checks to the `addSourceFile` processor to make sure the JSON file isn’t missing the #PLATFORM# or #DEVICE# token when the respective arrays contain more than one element.

### DIFF
--- a/tests/BrowscapTest/Data/DataCollectionTest.php
+++ b/tests/BrowscapTest/Data/DataCollectionTest.php
@@ -825,6 +825,34 @@ HERE;
     }
 
     /**
+     * checks if an exception is thrown if the devices entry has multiple items and there is no #DEVICE# token
+     *
+     * @expectedException \LogicException
+     * @expectedExceptionMessage the "devices" entry contains multiple devices but there is no #DEVICE# token for key
+     *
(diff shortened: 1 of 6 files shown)


tests/BrowscapTest/Data/DataCollectionTest.php
```diff
@@ -825,6 +825,34 @@ public function testAddSourceFileThrowsExceptionIfDevicesEntryIsNotAnArray()
     }
 
     /**
+     * checks if an exception is thrown if the devices entry has multiple items and there is no #DEVICE# token
+     *
+     * @expectedException \LogicException
```
asgrim: Just a nitpicking, but should use​ `expectException` method invocation instead of annotation (same with message) in future

PR #1294, opened 2017-03-20
Condensing devices for rest of yabrowser files

jaydiablo:
This wraps up the files in the browsers/yabrowser directory.


asgrim:
thanks @jaydiablo!

PR #1295, opened 2017-03-20
pattern cleanup

mimmi20:
This PR is related to browscap/browscap-php#179. Most of the useragents noted there are added as tests. Missing pattern were added and different writings in pattern are unified.

This should solve issue #1219.


resources/user-agents/apps/facebook/facebook-app.json
```diff
@@ -195,22 +195,21 @@
             "N9519": "ZTE N9519",
             "Nexus 5": "LG Nexus 5",
             "R1 HD": "BLU R1 HD",
-            "SAMSUNG-SM-G891A": "Samsung SM-G891A",
```
jaydiablo: Interesting, is this how we should deal with Samsung devices that prefix with `SAMSUNG-` in the future?

Any concern that the `#PLATFORM#` replacement won't end with an asterisk, or is it safe to assume that it always will?

resources/user-agents/apps/facebook/facebook-app.json
```diff
@@ -456,7 +455,13 @@
             "XT1080": "Motorola XT1080",
             "XT907": "Motorola XT907",
             "Z793C": "ZTE Z793C",
-            "Z970": "ZTE Z970"
+            "Z970": "ZTE Z970",
+            "GT-N5110": "Samsung GT-N5110",
```
jaydiablo: As you may have noticed in some of my recent PRs, I've been trying to alphabetize these device blocks (by device match text. I just highlight the block in Sublime and press F5), as it makes it easier to quickly scan the list to see if a device's match exists already or not.

Any objections in doing this moving forward?  I think it would be nice to eventually have a sort of cleanup tool that does these sorts of things automatically (unless you have something already).

resources/user-agents/apps/facebook/facebook-app.json
mimmi20: If you look into line 213 of this file, there was the same device without the Samsung-Prefix. I think, its easier to handle without that prefix.

resources/user-agents/apps/facebook/facebook-app.json
mimmi20: no objections

asgrim:
thanks @mimmi20!

PR #1302, opened 2017-03-25
Adding Firefox on iOS versions/platforms/devices

### LINKED ISSUE #1272: Firefox on iPad
see donatj/PhpUserAgent#40 for more details.

jaydiablo:
This fixes #1272.

In addition to the UA in the referenced ticket I also dug through our
server logs and collected a variety of Firefox on iOS useragents that
spanned different versions/iOS versions/devices.

This resulted in me adding a number of Firefox iOS versions, a few
platforms to the Firefox 1.x file, and iPod Touch as a device for both
Firefox iOS files.


jaydiablo:
FYI, here's the full list of unique Firefox iOS useragents I found in our logs, and how Browscap parses them with these changes in place:

[firefox-ios.csv.zip](https://github.com/browscap/browscap/files/870211/firefox-ios.csv.zip)

The only ones that don't have a proper Platform version are the ones that say they're iOS 10.3.  10.3 isn't out yet, but it's apparently coming soon:

https://appadvice.com/post/apple-ios-10-3-release/742971

https://developer.apple.com/library/prerelease/content/releasenotes/General/WhatsNewIniOS/Articles/iOS10_3.html

>This article summarizes the key developer-related features introduced in iOS 10.3, which runs on currently shipping iOS devices. 

FWIW, we are starting to see 10.3 iOS versions in our logs, but still a really low percentage (I assume this is people using the Beta version).

PR #1312, opened 2017-03-26
Adding iOS 10.3

### LINKED ISSUE #1303: Add iOS 10.3
I mentioned this in #1302, but we're starting to see iOS 10.3 versions in our logs (most likely users using the Beta version of the OS).

Sounds like it should be released pretty soon, but isn't out yet (outside of the beta program).

Should probably consider adding it soon no?

### LINKED ISSUE #1305: Puffin on iOS being detected as Safari
These useragents should probably be Puffin, not Safari:

>Mozilla/5.0 (iPad; CPU OS 10_3 like Mac OS X; de-AT) AppleWebKit/537.36 (KHTML, like Gecko)  Version/10.3 Mobile/14E5277a Safari/537.36 Puffin/5.2.0IT
Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X; en-US) AppleWebKit/537.36 (KHTML, like Gecko)  Version/10.3 Mobile/14E5230e Safari/537.36 Puffin/5.2.0IP
Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X; en-US) AppleWebKit/537.36 (KHTML, like Gecko)  Version/10.3 Mobile/14E5239e Safari/537.36 Puffin/5.2.0IP
Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X; ko-Kore-KR) AppleWebKit/537.36 (KHTML, like Gecko)  Version/10.3 Mobile/14E5249d Safari/537.36 Puffin/5.2.0IP
Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X; ru-RU) AppleWebKit/537.36 (KHTML, like Gecko)  Version/10.3 Mobile/14E5239e Safari/537.36 Puffin/5.2.0IP

This is against current master branch.

jaydiablo:
This fixes #1303. I’ve added iOS 10.3 in all places where 10.2 was
already set.

This also fixes #1305 since I needed to add 10.3 to match against the
UAs in that ticket, I just rolled the fix (version bump, platforms add
(the iOS_A variants), swap P at the end with ?) into this update.

I added tests with legit 10.3 useragents from our logs where possible.


resources/user-agents/browsers/puffin/puffin-5-0-on.json
```diff
@@ -101,24 +101,28 @@
...
         {
           "match": "Mozilla/5.0 (#PLATFORM#)*applewebkit*(*khtml*like*gecko*)*Version/*Safari/* Puffin/#MAJORVER#.#MINORVER#.*I?*",
           "device": "general Mobile Device (Apple)",
           "engine": "WebKit",
           "platforms": [
-            "iOS_A_10_0", "iOS_A_10_1", "iOS_A_10_2",
-            "iOS_A"
+            "iOS_A_10_0", "iOS_A_10_1", "iOS_A_10_2", "iOS_A_10_3",
+            "iOS_A",
+            "iOS_C_10_0", "iOS_C_10_1", "iOS_C_10_2", "iOS_C_10_3",
```
mimmi20: Please change the ordering for iOS_C.

resources/user-agents/browsers/puffin/puffin-5-0-on.json
```diff
@@ -101,24 +101,28 @@
...
             "iPhone": "iPhone",
             "iPad": "iPad"
           },
           "engine": "WebKit",
           "platforms": [
-            "iOS_A_10_0", "iOS_A_10_1", "iOS_A_10_2",
-            "iOS_A"
+            "iOS_A_10_0", "iOS_A_10_1", "iOS_A_10_2", "iOS_A_10_3",
+            "iOS_A",
+            "iOS_C_10_0", "iOS_C_10_1", "iOS_C_10_2", "iOS_C_10_3",
```
mimmi20: Please change the ordering for iOS_C.

resources/user-agents/browsers/puffin/puffin-5-0-on.json
jaydiablo: I've noticed that the C's are typically ordered highest to lowest and the A's are ordered lowest to highest, why is that?

mimmi20: iOS_C_8_0 is coded without the zero, iOS_C_9_0 too. Because of this iOS 8 and 9 must have their Zero version at last. iOS_C_10_0 is coded with the Zero, buit we should keep the ordering like in the other places.
Maybe we should change the ordering for iOS_A in a separate PR too.

mimmi20: I dont really remember why I started this.

jaydiablo: When I've been consolidating Android devices in some of the larger files, I've been rearranging the Android platform versions to be highest -> lowest, as I find it easier to scan the file and sort the child groups.

Does the lack of the zero, and the order of the pattern matter if the lack of a zero causes the pattern to be shorter anyhow (which browscap-php will move down a notch in the "priority" list when parsing)?

jaydiablo: If that seems to make sense across the board (highest to lowest for platform versions and browser versions) then I think we should change the iOS_A_* definitions to be ordered that way as well, assuming that doesn't break any tests.

mimmi20: This should have no effect for browscap-php, but maybe for other clients, but normally we dont care about them.

Some time ago (maybe before browscap-php v3) I wanted to change the sorting in the ini file, that the client does not need to resort anything. But this did not work.

asgrim:
I :heart: that codecov report @jaydiablo :grin:

PR #1329, opened 2017-04-02
allow Edge 12 without Build Number

### LINKED ISSUE #1318: Edge user agent is detected as Chrome
see ua-parser/uap-core#233 for more details.

mimmi20:
I removed the "." from the pattern for Edge 12. This fixes #1318.


jaydiablo:
It looks like the reason the codecov report changed to 87.11% is because the coverage.json file was never built because there was a warning in the tests:

https://travis-ci.org/browscap/browscap/jobs/217812810#L507

Looks like you named the test case `issue-1313-A` but it should be `issue-1318-A`.  This caused the Full test to abort, which prevented the coverage from being built (since the collected pattern ids were empty).

tests/fixtures/issues/issue-1318.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+return [
+    'issue-1313-A' => [
```
jaydiablo: Should be `issue-1318-A`.

jaydiablo:
Oh, I just wanted to confirm that this change won't cause an issue if/when Edge reaches version 120.  Initially, I suppose that version will detect as version 12, but once 120 is added, that should clear that up right (seems unlikely it'll reach that version though ;) ).

mimmi20:
Only version 12 is detected without ".". All other versions have a "." after the major version. If Edge reaches version 120 it will be "120.*" in our pattern.

Issue #224, opened 2014-04-23
new user agents


gocher68:
Mozilla/5.0 (Macintosh; Intel Mac OS X 1084) AppleWebKit/536.29.13 (KHTML like Gecko) Version/6.0.4 Safari/536.29.13
Mozilla/5.0 (Macintosh; Intel Mac OS X 107) AppleWebKit/534.48.3 (KHTML like Gecko) Version/5.1 Safari/534.48.3

PR #1331, opened 2017-04-03
Cleaning up Safari/Safari Webview Detection

### LINKED ISSUE #1323: New Safari Version + iOS 10 for Mobile Safari Web View
Saw this today: https://webkit.org/blog/7477/new-web-features-in-safari-10-1/

Looks like a new Safari version for desktop mac OS (10.1).

I also noticed that the mobile Safari UI Web view is lacking the iOS 10.x versions.

I have a PR nearly ready for this, just wanted an issue number to name the test case.

jaydiablo:
This fixes #1323 but does a lot more than that (sorry).

I first started this by bumping the versions necessary for #1323, but
then decided to pull a bunch of useragents from our logs and run them
through the changes. This exposed a lot of useragents that didn’t seem
like they were parsing correctly either due to missing
versions/platform versions or the entire browser itself.

I did as much research into these changes as I could, including
downloading different browsers to test if the useragents found in our
logs really do coincide to how the actual browser reports.

I’ll add some inline comments describing why I made the changes I did.
I tried to include a test case for anything new that I added (within
reason, I didn’t try to have a new test case for every version/platform
version though) so coverage should still be pretty good, but I expect
it’ll drop since a lot of new platform versions were added.


resources/user-agents/browsers/misc/smarttv-webbrowser.json
```diff
@@ -29,6 +30,72 @@
           }
         }
       ]
+    },
+    {
+      "userAgent": "SmartTV WebBrowser Webkit",
```
jaydiablo: These patterns come from the `safari-x-.json` files.  I looked around a bit to see if what these TVs use is actually safari, which doesn't seem to be the case. It sounds like it's definitely a webkit based browser, but isn't necessarily "Safari" since it's an embedded OS.  I thought it was misleading to keep it as "Safari" so moved them over to this file to name them "SmartTV WebBrowser" which I think is more indicative of what they are.

resources/user-agents/browsers/mobile-safari-uiwebview/mobile-safari-uiwebview.json
```diff
@@ -24,71 +24,67 @@
...
             "iOS_B",
             "iOS_C_4_2", "iOS_C_4_3", "iOS_C_5_0", "iOS_C_5_1", "iOS_C_6_1", "iOS_C_6_0", "iOS_C_7_1", "iOS_C_7_0",
             "iOS_C_8_4", "iOS_C_8_3", "iOS_C_8_2", "iOS_C_8_1", "iOS_C_8_0", "iOS_C_9_3", "iOS_C_9_2", "iOS_C_9_1", "iOS_C_9_0",
+            "iOS_C_10_3", "iOS_C_10_2", "iOS_C_10_1", "iOS_C_10_0",
             "iOS_C"
           ]
         },
         {
-          "match": "Mozilla/5.0*(#PLATFORM#)*applewebkit*(*khtml*like*gecko*)*Mobile*",
+          "match": "Mozilla/5.0 (#DEVICE##PLATFORM#)*",
```
jaydiablo: I added this pattern as I did see a lot of agents like this getting caught by the below, more generic patterns.  I thought it was helpful to have the iOS version detection here when the patterns can match this format.

resources/user-agents/browsers/mobile-safari-uiwebview/mobile-safari-uiwebview.json
```diff
@@ -24,71 +24,67 @@
...
-          "device": "iPad"
-        },
-        {
-          "match": "*iPhone*",
-          "device": "iPhone"
-        },
-        {
-          "match": "*iPod*",
-          "device": "iPod Touch"
+          "match": "*#DEVICE#*",
```
jaydiablo: We discussed this some in #1323, ultimately I decided to just keep all of these generic patterns, just condensed them.  I think leaving them is fine, any agents that seem like they should be brought to their own file can be found easily even with these patterns in place.

resources/user-agents/browsers/mobile-safari/mobile-safari-3-x.json
```diff
@@ -28,8 +28,11 @@
       },
       "children": [
         {
-          "match": "Mozilla/5.0 (iPhone#PLATFORM#)*applewebkit*(*khtml*like*gecko*)*Version/#MAJORVER#.#MINORVER#*Mobile/*Safari/*",
-          "device": "iPhone",
+          "match": "Mozilla/5.0 (#DEVICE##PLATFORM#)*applewebkit*(*khtml*like*gecko*)*Version/#MAJORVER#.#MINORVER#*Mobile/*Safari/*",
```
jaydiablo: Just condensing here.

resources/user-agents/browsers/mobile-safari/mobile-safari-8-x.json
```diff
@@ -35,27 +35,35 @@
             "iPod": "iPod Touch"
           },
           "platforms": [
-            "iOS_A_6_0", "iOS_A_6_1", "iOS_A_7_0", "iOS_A_7_1", "iOS_A_8_0", "iOS_A_8_0_B",
```
jaydiablo: There's no change of versions here, I just was re-organizing a bit to get like-Major versions on the same line.

resources/user-agents/browsers/opera-mini/opera-mini-8-0-on.json
```diff
@@ -139,24 +139,6 @@
           "platforms": [
             "JAVA"
           ]
-        },
-        {
```
jaydiablo: These got moved to their own `opera-mini-ios.json` file because I felt there were too many versions in this file (Opera Mini iOS is only at version 14 right now: https://itunes.apple.com/app/id363729560?mt=8&pt=341230&ct=ose_google_via_opera_com_mobile_mini_iphone_top)

resources/user-agents/browsers/safari/safari-10-0-on.json
```diff
@@ -30,15 +30,15 @@
           "match": "Mozilla/5.0 (#PLATFORM#) applewebkit* (*khtml*like*gecko*) *Version/#MAJORVER#.#MINORVER#* Safari/*",
           "device": "Macintosh",
           "platforms": [
-            "OSX_10_12", "OSX_10_11", "OSX_10_10",
+            "OSX_10_12", "OSX_10_11", "OSX_10_10", "OSX_10_9",
```
jaydiablo: So, for the Safari on OSX platform versions I used this wikipedia article: https://en.wikipedia.org/wiki/Safari_version_history#Mac to try to limit the OSX versions to just what's in that table, but to be a bit conservative, I went +1 and -1 of the versions in that table to account for pre-release versions and perhaps cases where a new version of OSX would be using the previous version of Safari (this does seem like it would be rare though).

Open to discussion about it. Like the iOS version stuff I mentioned earlier, this won't cause any "Default Browsers" situations, as the generic platforms will still catch the agent, it just won't have a platform version.

resources/user-agents/browsers/safari/safari-4-0.json
```diff
@@ -27,10 +27,10 @@
       },
       "children": [
         {
-          "match": "Mozilla/5.0 (SMART-TV; X11;#PLATFORM#) applewebkit* (*khtml*like*gecko*) *Version/#MAJORVER#.#MINORVER#* Safari/*",
```
jaydiablo: As mentioned before, this moved to the `smartv-webbrowser.json` file.

resources/user-agents/browsers/safari/safari-4-0.json
```diff
@@ -27,10 +27,10 @@
...
-          "device": "Smart-TV",
+          "match": "Mozilla/5.0 (#PLATFORM#)*applewebkit*(*khtml*like*gecko*)*Version/#MAJORVER#.#MINORVER#*Safari/*",
```
jaydiablo: I did add Safari 4.1 on linux here as it seems possible to install Safari on linux using Wine, the caveat is that it has to be the Windows version of Safari, which only had a few releases (3.0 - 5.1): https://en.wikipedia.org/wiki/Safari_version_history#Windows

resources/user-agents/browsers/safari/safari-5-2.json
```diff
@@ -27,43 +27,26 @@
...
-        },
-        {
-          "match": "Mozilla/5.0 (#PLATFORM#) applewebkit* (*khtml*like*gecko*) *Version/#MAJORVER#.#MINORVER#* Safari/*",
-          "device": "Linux Desktop",
-          "platforms": [
-            "CentOS", "Linux_64", "Linux_32", "Linux"
-          ]
-        },
-        {
-          "match": "Mozilla/5.0 (#PLATFORM#) applewebkit* (*khtml*like*gecko*) *Version/#MAJORVER#.#MINORVER#* Safari/*",
```
jaydiablo: Safari 5.2, according to Wikipedia, wasn't released for Windows.

resources/user-agents/browsers/safari/safari-5-2.json
```diff
@@ -27,43 +27,26 @@
...
-        {
           "match": "Mozilla/5.0 (#PLATFORM#) applewebkit* (*khtml*like*gecko*) *Version/#MAJORVER#.#MINORVER#* Safari/*",
           "device": "Macintosh",
           "platforms": [
-            "OSX_10_4", "OSX_10_5", "OSX_10_6", "OSX_10_7", "OSX_10_8", "OSX_10_9", "OSX_10_10",
+            "OSX_10_9", "OSX_10_8", "OSX_10_7", "OSX_10_6",
             "OSX"
-          ]
+          ],
+          "properties": {
```
jaydiablo: It also wasn't released for Mac according to wikipedia, but I did find this article that discusses it being released as a Beta: https://www.macrumors.com/2012/02/16/apple-seeds-safari-5-2-for-os-x-lion-to-developers/ So I left it in and just marked it as a beta.

resources/user-agents/browsers/safari/safari-9-x.json
```diff
@@ -1,6 +1,6 @@
 {
   "division": "Safari #MAJORVER#.#MINORVER#",
-  "versions": ["9.1", 9, "8.1"],
```
jaydiablo: Removed 8.1 since it didn't show as being released in that wikipedia article (and no tests test for it).

resources/user-agents/browsers/safari/safari-generic.json
```diff
@@ -257,6 +250,146 @@
           ]
         }
       ]
+    },
+    {
+      "userAgent": "Safari Generic Older",
```
jaydiablo: I did look around a bit to see what Safari 1.x and early 2.x useragents looked like and found these examples, so added them here as special cases.  I thought about creating a `safari-1-x.json` file, but then found some patterns for early 2.x versions of Safari that needed a home as well.

resources/user-agents/browsers/sleipnir/sleipnir-ios-1-x.json
```diff
@@ -0,0 +1,59 @@
+{
```
jaydiablo: I broke these Sleipnir versions into separate files due to the minimum iOS version bump that the major versions had (based on archive.org captures of the iTunes Store page for this app).

resources/user-agents/browsers/yabrowser/yabrowser-15-on.json
```diff
@@ -315,28 +315,37 @@
...
+            "iOS_C_7_1", "iOS_C_7_0",
+            "iOS_C_8_3", "iOS_C_8_2", "iOS_C_8_1", "iOS_C_8_0",
             "iOS_C_9_3", "iOS_C_9_2", "iOS_C_9_1", "iOS_C_9_0",
             "iOS_C_10_3", "iOS_C_10_2", "iOS_C_10_1", "iOS_C_10_0",
             "iOS_C"
           ]
         },
         {
-          "match": "Mozilla/5.0 (#PLATFORM#) AppleWebKit/* (KHTML* like Gecko) YaBrowser/#MAJORVER#.#MINORVER#* Safari/*",
+          "match": "Mozilla/5.0 (#PLATFORM#) AppleWebKit/* (*khtml*like*gecko*) YaBrowser/#MAJORVER#.#MINORVER#* */*",
```
jaydiablo: There's a test for this, I saw some useragents that have `YaApp_iOS` in the spot where `Safari` is supposed to be.

src/Browscap/Generator/Helper/BuildHelper.php
```diff
@@ -160,7 +160,9 @@ public static function run(
 
                 foreach (array_keys($sectionsWithVersion) as $sectionName) {
                     if (array_key_exists($sectionName, $output)) {
-                        $logger->error('tried to add section "' . $sectionName . '" more than once -> skipped');
+                        $logger->error(
+                            'tried to add section "' . $sectionName . '" from "' . $division->getName() . '" more than once -> skipped'
```
jaydiablo: I added this to help find what pattern was causing this error, figured it wouldn't hurt to keep it.

src/Browscap/Generator/Helper/BuildHelper.php
```diff
@@ -177,7 +179,7 @@ public static function run(
                         ->renderSectionBody($section, $collection, $sectionsWithVersion, $sectionName)
                         ->renderSectionFooter($sectionName);
 
-                    $output[$sectionName] = $sectionName;
+                    $output[$sectionName] = 1;
```
jaydiablo: This `$output` variable is only ever used on line 162 to see if the array key exists. I figured it would save a little bit of memory to not store the same value as the key and value in the array if only the key is ever used.

tests/fixtures/issues/issue-224.php
```diff
@@ -1,59 +1,6 @@
 <?php
 
 return [
-    'issue-224-A' => [
```
jaydiablo: This agent was giving me a bit of grief with some of the changes I made, then I looked it up in the issue tracker: https://github.com/browscap/browscap/issues/224

Turns out it was a fake agent that was later converted to a regular test when the fakes were removed in this commit: https://github.com/mimmi20/browscap/commit/d0869644bc438539e4bc14b7084b43ea1d4c50f0

The OSX platform here gets detected as "10.4" (from the string `1084`) which isn't correct (if anything it should be "10.8").  With my changes, Safari 6.0 isn't available on OSX 10.4. 

I decided to remove the test case since the agent isn't legit (I think the one below should probably removed too, but it's less of an issue).

tests/fixtures/issues/issue-523.php
```diff
@@ -1079,7 +1079,7 @@
             'Platform_Bits' => '32',
             'Platform_Maker' => 'Apple Inc',
             'Alpha' => false,
-            'Beta' => true,
+            'Beta' => false,
```
jaydiablo: Opera mini is marked as a beta, but when I moved the iOS patterns to their own file, I removed that property as there doesn't seem to be any indication that Opera Mini is beta on iOS: https://itunes.apple.com/app/id363729560?mt=8&pt=341230&ct=ose_google_via_opera_com_mobile_mini_iphone_top

mimmi20:
@jaydiablo Could you solve the conflicts?

jaydiablo:
@mimmi20 conflicts resolved

PR #1336, opened 2017-04-04
Locking version for browscap-php, adding composer validate to travis

jaydiablo:
Fixes #1335.

I added the composer validate command to prevent the composer.lock file
from getting out of sync with composer.json.

PR #1337, opened 2017-04-05
Adding "failOnWarning" to phpunit.xml

### LINKED ISSUE #1330: Tests should fail if Data Provider is invalid
In https://github.com/browscap/browscap/pull/1329 I noticed that the coverage report for the patterns wasn't uploaded: 

https://travis-ci.org/browscap/browscap/jobs/217812810#L560

Which is because the Full useragent tests didn't run due to this warning:

https://travis-ci.org/browscap/browscap/jobs/217812810#L507

I'd expect the tests to fail in this case, rather than silently ignore the warning. Not sure if this is related to this PHPUnit issue (which was closed as invalid):

https://github.com/sebastianbergmann/phpunit/issues/2449

Or just need to have a setting set to make sure that Warnings cause the tests to fail.

jaydiablo:
Fixes #1330.

By default, it seems, PHPUnit does not return a > 0 exit status code
for tests that throw warnings:

```
There was 1 warning:

1) Warning
The data provider specified for
BrowscapTest\UserAgentsTest::testUserAgentsFull is invalid.
Test data is duplicated for key "issue-004"

WARNINGS!
Tests: 22, Assertions: 234, Warnings: 1.
MacBook-Pro:browscap jasonklehr$ echo $?
0
```

Compared to when there is a failure:

```
FAILURES!
Tests: 15580, Assertions: 801930, Failures: 2.
MacBook-Pro:browscap jasonklehr$ echo $?
1
```

By specifying this config in the XML, the exit status will change for
test runs that throw warnings:

```
There was 1 warning:

1) Warning
The data provider specified for
BrowscapTest\UserAgentsTest::testUserAgentsFull is invalid.
Test data is duplicated for key "issue-004"

WARNINGS!
Tests: 22, Assertions: 234, Warnings: 1.
MacBook-Pro:browscap jasonklehr$ echo $?
1
```

Which should fail the travis build.  This could be specified on just
the CLI, just for a specific run (like the useragent test run), but I
don’t think there really should be any test runs that are ok to be
throwing warnings (especially since they don’t currently).

NOTE, this is failing travis for 2 reasons (both related to the same issue).

1) master branch of browscap-php has php 7.1 syntax, so any builds that use php 7.0 are failing (which is all builds except the nightlies and the useragent test build).

2) changes in master branch of browscap-php make some classes final, which means they can’t be mocked, which is throwing a warning in a test in a nightly build (this may happen in regular builds of 7.1 too, but right now the only 7.1 build is the useragent test, which doesn’t run those tests).

Both of these (I believe) are fixed with #1336.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,6 +9,7 @@
     verbose="true"
     colors="auto"
     bootstrap="vendor/autoload.php"
+    failOnWarning="true"
     >
     <php>
         <ini name="error_reporting" value="-1"/>


phpunit.xml.dist
```diff
@@ -9,6 +9,7 @@
     verbose="true"
     colors="auto"
     bootstrap="vendor/autoload.php"
+    failOnWarning="true"
```
mimmi20: Isn't this true by default?

jaydiablo: Not based on my tests (as seen above). When the setting isn't set, PHPUnit does complain about the warning, but the exit code is `0` (success), which is what we saw happen in #1318 (the travis build succeeded even though the warning was encountered).  With that config option set (or the `--fail-on-warning` CLI flag is set) the PHPUnit complains **and** sets the exit code to `1`.

Here's what I can find in the PHPUnit source:

https://github.com/sebastianbergmann/phpunit/blob/e083cb6682159fcd7863464729211aae22c1818a/src/Util/Configuration.php#L700

https://github.com/sebastianbergmann/phpunit/blob/e083cb6682159fcd7863464729211aae22c1818a/src/TextUI/Command.php#L109

https://github.com/sebastianbergmann/phpunit/blob/e083cb6682159fcd7863464729211aae22c1818a/src/TextUI/Command.php#L510

PR #1346, opened 2017-04-08
Cleaning up iTunes detection

### LINKED ISSUE #1345: Missing iTunes agents
This is part of #1344.

I noticed a number of iTunes agents that were being parsed as "Default Browser".

The PR for this will have the agents in a test file (just need an issue number).

jaydiablo:
Fixes #1345

Noticed that the existing iTunes detection wasn’t catching some patterns in our logs for March.

This adds those patterns plus some more I found when digging further in our logs, to the tests.

I moved the itunes.json file out of the apple folder in the apps section in order to be able to add more files covering the various versions of itunes on desktop and on ios.

I separated desktop and ios because there isn’t a lot of version overlap (at least for useragents I found in our logs), so it would’ve been wasteful to have the ios patterns expanding in the desktop versions.

For the desktop patterns, I used this table: https://en.wikipedia.org/wiki/History_of_iTunes#Operating_system_versions to determine which versions of OSX should be included in each version range.  This should help reduce the number of patterns, and also help identify platform/browser version mismatches.

I also noticed, that after running the tests and looking at the coverage, that the more recent versions of iTunes use the `OSX_B_*` style platform strings but older versions use the `OSX_*` style, so I removed the unnecessary ones.

This does also add a generic “Windows 64” platform that I’m using in these iTunes patterns for windows desktop to differentiate the 64 bit versions of windows.


resources/user-agents/apps/itunes/itunes-1-2-to-4-7.json
```diff
@@ -0,0 +1,55 @@
...
+        },
+        {
+          "match": "iTunes/#MAJORVER#.#MINORVER#* (#PLATFORM#)*",
+          "device": "Windows Desktop",
+          "platforms": [
+            "Windows", "Windows_64"
+          ]
+        },
+        {
+          "match": "iTunes?#MAJORVER#.#MINORVER#*",
```
jaydiablo: I had to use the question mark here on these generic patterns because otherwise these patterns were catching the Downcast and PodCruncher agents in the tests, which look like this:

>iTunes/10.7 Downcast/5608
>iTunes/10.7 Downcast/2.8.14.1002
>iTunes/10.5.2 (PodCruncher 2.2)

The patterns for these look like this:

```
iTunes/* Downcast/2.9*
iTunes/* Downcast/2.8*
iTunes/* Downcast/*
iTunes/* (PodCruncher *
```

The asterisk after the slash would cause these patterns to lose priority to the version expanded iTunes pattern from these files, which would look something like this after expansion:

```
iTunes/4.7*
```

Which is longer before the first wildcard character.  This is related to the issues described in https://github.com/browscap/browscap-php/issues/179

The question mark reduces the priority on these. Perhaps not ideal (since it'll match other varieties of this, like `iTunes_4.8`, but without fixing the parser I don't think there's another way to deal with this.

Issue #1344, opened 2017-04-08
Missing user-agents


jaydiablo:
I'm going through our logs for March and pulling out any useragents that parsed as "Default Browser" and trying to decipher the valid ones.

I want to break it out in to a series of tickets/PRs rather than just one huge one, so using this as a meta issue of sorts (though I'll probably make an issue for a lot of the smaller/obscure browsers and attach them to this issue number).

You can assign this to me.

WhiteLoti:
Browscap doesn't seem to detect the linux 64 bit desktop user agent string "Mozilla/5.0 (X11; Linux x86_64; rv:52.0) Gecko/20100101 Firefox/52.0"

mimmi20:
@WhiteLoti Did you update your browscap.ini file to the newest version?

WhiteLoti:
Yes I downloaded it about 20 minutes ago and cached it.

jaydiablo:
@WhiteLoti the useragent appears to parse fine for Browscap 6022 (the latest version):

http://useragent.mkf.solutions/?userAgent=Mozilla%2F5.0+%28X11%3B+Linux+x86_64%3B+rv%3A52.0%29+Gecko%2F20100101+Firefox%2F52.0

Can you confirm that you're using version 6022?

PR #1348, opened 2017-04-08
Adding Android version of Google App

### LINKED ISSUE #1347: Google Search App on Android being detected as Android Browser/Webview
This is part of #1344.

Browscap has the iOS version of the Google Search App (GSA) pretty well covered, but seems to be lacking the Android version of the app.

I have a PR just about ready for this (have to wrap up the tests).

jaydiablo:
Fixes #1347 

This adds an Android version of the Google Search App to go with the iOS version.

I had to add devices to this in order to make sure these particular user-agents were detected as the GSA app instead of Android Browser or Android Webview (since the root of the pattern is very similar, and the only difference is the presence of the "GSA/" string. "GSA/" alone isn't enough to make these patterns longer than the Android Browser/Webview patterns that contain device info).

Of course, there are other devices in the Android Browser/Webview patterns that would prevent some agents from being detected as GSA, I just didn't encounter those ones in the list of agents I pulled to test this.

Speaking of that, I did pull some ~2000 useragents from our logs that had that "GSA/" pattern and ran these changes against that list. All of them were detected as "Google App" instead of Android Browser/Webview, as expected.


resources/devices/mobile-phone/samsung.json
```diff
@@ -5628,6 +5628,19 @@
             "isTablet": "false"
         }
     },
+    "Samsung SM-G530BT": {
```
jaydiablo: This one and the one below aren't used in any changes in this diff, but some other changes coming as part of #1344. Since I had made some other changes to this file, I just added those in now.

resources/devices/tablet/lenovo.json
```diff
@@ -1,4 +1,17 @@
 {
+    "Lenovo 1838": {
```
jaydiablo: Same for this device.

jaydiablo:
Hmm, that test failure is odd.  All I did in that commit was remove those two new devices which aren't used by any useragents in this PR (they're added back in in #1367).  Tests pass locally for this branch without any uncommitted changes.

mimmi20:
@jaydiablo Could you check the failing test?

jaydiablo:
@mimmi20 I can't reproduce the failure locally, but it does seem that issue-885-CN should be updated to address the changes made to the samsung device file.

I'm not sure why the test passes locally (and passed on travis earlier).  I'll update it.

jaydiablo:
Oh, duh, issue-885 was added recently (by me, hah!).

jaydiablo:
@mimmi20 all green, again :)

PR #1351, opened 2017-04-09
Adding some devices/versions of Firefox on FirefoxOS

### LINKED ISSUE #1350: Some missing devices/versions for Firefox on Firefox OS
This is part of #1344.

Noticed some devices/firefox versions missing.

jaydiablo:
Fixes #1350

Found some patterns that weren’t matching to FirefoxOS like they should
have been.

I added Firefox OS 2.5, which I’ve marked as Beta. Looks like it was
the next planned release of Firefox OS, but the project was cancelled
before it was released. I assume that the versions of Firefox on the
OPENC phone I’m seeing are due to this beta release, but they could be
fake.

https://en.wikipedia.org/wiki/Firefox_OS#Release_history


resources/user-agents/browsers/firefox-on-firefoxos/firefox-os-1-1.json
```diff
@@ -28,36 +28,30 @@
...
+            "ALCATELOneTouch4012X*": "Alcatel OT-4012X",
+            "ALCATEL ONE TOUCH 4012A": "Alcatel OT-4012A",
+            "ALCATELOneTouch4012A": "Alcatel OT-4012A",
+            "LG-D300": "LG D300",
+            "ZTEOPEN": "ZTE Open"
+          }
         },
         {
-          "match": "Mozilla/5.0 (Mobile; rv:#MAJORVER#.#MINORVER#*)*Gecko/#MAJORVER#.#MINORVER#*Firefox/#MAJORVER#.#MINORVER#*"
+          "match": "Mozilla/5.0 (Mobile; *rv:#MAJORVER#.#MINORVER#*)*Gecko/#MAJORVER#.#MINORVER#*Firefox/#MAJORVER#.#MINORVER#*"
```
jaydiablo: I added the asterisk here to allow for these patterns to catch other devices that aren't in browscap yet, so at least they're being caught as Firefox on Firefox OS instead of Default Browser.

There aren't a lot of devices that are compatible with Firefox OS, but without seeing the agents in the wild, I don't know what their device identifiers are.

https://en.wikipedia.org/wiki/Comparison_of_Firefox_OS_devices

tests/fixtures/issues/issue-492.php
```diff
@@ -5103,7 +5103,7 @@
             'MinorVer' => '0',
             'Platform' => 'FirefoxOS',
             'Platform_Version' => 'unknown',
-            'Platform_Description' => 'Android OS',
+            'Platform_Description' => 'FirefoxOS',
```
jaydiablo: This one was sort of unexpected. When I added the platform description to FirefoxOS I expected all of the "unknown"s to need replacement, but not this one.

It must be some sort of inheritance related issue.

PR #1374, opened 2017-04-14
Wrapping up missing user-agents

### LINKED ISSUE #1344: Missing user-agents

jaydiablo:
This fixes #1344.

Adding the rest of the miscellaneous useragents that I didn’t make a separate issue for.

Most of these required new platform or versions for existing browsers, but some of the more obscure ones required adding new browsers.  I’ll add some inline comments to explain some of the changes.


resources/user-agents/browsers/iridium/iridium.json
```diff
@@ -1,6 +1,6 @@
 {
   "division": "Iridium Browser #MAJORVER#.#MINORVER#",
-  "versions": [44, "43.4"],
```
jaydiablo: All of the patterns in this file only have `#MAJORVER#` in the pattern, so it doesn't make sense to have a minor version here.

This was passing the test (which I've changed) since there are no other "43.x" versions here, but it would end up treating all "43.x" string as "43.4".

tests/fixtures/issues/issue-1059.php
```diff
@@ -2389,15 +2389,15 @@
     'issue-1059-64' => [
         'ua' => 'Mozilla/5.0 (X11; OpenBSD amd64) AppleWebKit/537.36 (KHTML, like Gecko) Iridium/43.4 Safari/537.36 Chrome/43.0.2357.125',
         'properties' => [
-            'Comment' => 'Iridium Browser 43.4',
```
jaydiablo: Here's the test case I modified for that version change in the iridium file.

tests/fixtures/issues/issue-1253.php
```diff
@@ -6,7 +6,7 @@
         'properties' => [
             'Comment' => 'Kindle',
             'Browser' => 'Kindle',
-            'Browser_Type' => 'unknown',
+            'Browser_Type' => 'Browser',
```
jaydiablo: Figured the Kindle UAs should probably be a Browser, so updated that and had to modify these tests to match.

PR #1375, opened 2017-04-14
Condensing devices in chrome-11.json

jaydiablo:
Several dupe, or near dupe patterns here.

### DIFF
--- a/resources/user-agents/browsers/chrome/chrome-11.json
+++ b/resources/user-agents/browsers/chrome/chrome-11.json
@@ -29,119 +29,79 @@
       },
       "children": [
         {
-          "match": "Mozilla/5.0 (X11; Linux*; HTC/Sensation/*) applewebkit* (*khtml*like*gecko*) Chrome/#MAJORVER#.*Safari/*",
-          "device": "HTC Z710",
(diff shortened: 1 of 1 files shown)


resources/user-agents/browsers/chrome/chrome-11.json
```diff
@@ -29,119 +29,79 @@
...
       "children": [
         {
-          "match": "Mozilla/5.0 (X11; Linux*; HTC/Sensation/*) applewebkit* (*khtml*like*gecko*) Chrome/#MAJORVER#.*Safari/*",
-          "device": "HTC Z710",
-          "platforms": [
-            "Android_4_0"
-          ]
-        },
-        {
-          "match": "Mozilla/5.0 (X11; Linux x86_64*; HTC/Sensation/*) applewebkit* (*khtml*like*gecko*) Chrome/#MAJORVER#.*Safari/*",
```
jaydiablo: This one is a near dupe of the one above, just has the "x86_64" bit added.  Since there was no platform change or anything with that addition, I felt it was unnecessary to have both of these, since the one above should work for both here.

PR #1379, opened 2017-04-14
Adding some more useragents that weren't detected

### LINKED ISSUE #1378: Some more missing user-agents
With #1344 complete, I ran browscap against a list I had of undetected agents from February.  Most of the legit ones are detected now, but there is still a handful that I'd like to add patterns for.

Creating this issue so I have an issue number to attach them to.

jaydiablo:
Fixes #1378.

These all (except a couple, which I’ll explain in some inline comments) comes from our logs for February, as explained in #1378.

Some new patterns, some updates to existing patterns.


resources/platforms.json
```diff
@@ -1689,6 +1689,15 @@
         "Platform_Version": "11.04"
       }
     },
+    "Ubuntu_11_10": {
```
jaydiablo: I initially ran into a useragent that was Ubuntu 11.10 for Chromium, so I added this in anticipation of dealing with that agent. The agent ended up not being legit (was missing the `Chrome/*` part), but I figured I'd keep this here anyhow. I did fish out a Chromium Ubuntu 11.10 useragent from our logs though and added it to the test cases.

resources/user-agents/browsers/boxee/boxee.json
```diff
@@ -7,6 +7,7 @@
     {
       "userAgent": "Boxee",
       "device": "Boxee",
+      "platform": "Linux",
```
jaydiablo: All the patterns here were defined as "Linux" platform, so I just moved it up here.  I had to leave it for the one that has `#PLATFORM#` in the match though.

resources/user-agents/browsers/chromium/chromium-28-39.json
```diff
@@ -47,13 +48,13 @@
           ]
         },
         {
-          "match": "Mozilla/5.0 (*Linux i686*) applewebkit* (*khtml*like*gecko*) Ubuntu Chromium/#MAJORVER#.*Chrome/*Safari/*",
+          "match": "Mozilla/5.0 (*Linux i686*) applewebkit* (*khtml*like*gecko*) Ubuntu* Chromium/#MAJORVER#.*Chrome/*Safari/*",
```
jaydiablo: I added the asterisk here so that these patterns will pick up Ubuntu useragents and mark them as Ubuntu, rather than the generic "Linux" (They typically look like `Ubuntu/11.04`).

It may be nice to add more Ubuntu versions and add them to the pattern above, but this should at least catch other versions of Ubuntu that aren't explicitly defined.

resources/user-agents/browsers/chromium/chromium-55-on.json
```diff
@@ -65,6 +66,14 @@
           ]
         },
         {
+          "match": "Mozilla/5.0 (#PLATFORM#) applewebkit* Chromium/#MAJORVER#.*Safari/*",
```
jaydiablo: I only added this here since it's the only case I saw, though, I should add it to the generic division at least.

resources/user-agents/browsers/sundance/sundance-generic.json
```diff
@@ -0,0 +1,44 @@
...
+        "Frames": "true",
+        "IFrames": "true",
+        "Tables": "true",
+        "Cookies": "true",
+        "JavaScript": "true",
+        "CssVersion": "3"
+      },
+      "children": [
+        {
+          "match": "Sundance/* (Compatible; Windows; U; en-US)Version/*",
```
jaydiablo: This one comes from the useragent reported by the Sundance browser after downloading it to a Windows VM I have (http://www.digola.com/sundance.html).

The pattern below I made from an entry in our logs (which a couple of parsers also parse as Sundance). I'm not 100% sure that both of these are the same browser or not. Perhaps early versions of the browser used the UA below, but more recent versions use this one.

PR #1383, opened 2017-04-20
Cleaning up Firefox patterns, adding a lot of test coverage

### LINKED ISSUE #1381: Firefox platform versions
I was looking for useragents in our logs that had a browser name, but were lacking a browser version. The browser that this seemed to happen to the most was Firefox, so I extracted just Firefox useragents that didn't have a version to dig around a bit to see if they should have a version (and planning to make necessary changes so they get the proper version).

Which brings me to my question.  This has sort of come up before when dealing with Samsung browser (and whether or not to add versions of the browser that apparently haven't been released), and with Safari and OSX versions.

I was curious which versions of Firefox supported which versions of different platforms. It seems that some of this research has been done already, as I noticed that the platforms change a bit as you progress through Firefox versions, but it doesn't seem to be quite exact.  For example, Firefox 2.0 is the last version of Firefox that will run on Windows 98 (officially, more on this in a bit), here's the ticket where Windows 9x support was removed: https://bugzilla.mozilla.org/show_bug.cgi?id=330276

However, Browscap currently has Windows 98 as a platform for Firefox up to, and including version 3.8. There are 2 tests in the test suite which test for this (which I added as part of #1360, before doing any of this research).

Above I said "officially", and the reason for that is that there are projects out there that patch older versions of windows so that they'll be able to run newer versions of software, like browsers. This project is one such project and it allows Windows 98 users to run newer Firefox versions: http://kernelex.sourceforge.net/wiki/Mozilla_Firefox (though it doesn't seem to work reliably for all versions of Firefox, that wiki page claims that Firefox 3.6 runs pretty well, Firefox 10 should run pretty well, but anything else suffers from bugs/glitches, or may not run at all).

Most useragent parsers don't seem to care too much about this, since they're just looking for certain patterns in the string to determine which OS/version and Browser/version the agent is. Few look at all of the properties together when deciding how to parse.  For example, here's a string I crafted that has Firefox 42 running on Windows 98, which probably isn't possible even with the very best kernel extensions. :)

http://useragent.mkf.solutions/?userAgent=Mozilla%2F5.0+%28Windows%3B+U%3B+Win98%3B+bg%3B+rv%3A1.9.0.19%29+Gecko%2F20090624+Firefox%2F42.0

So my question is, how should Browscap behave in these situations?  Aside from the parsing aspect, removing some platforms from versions of the browser that aren't supported would reduce the number of patterns in the Browscap file, which I think is a good thing.

As for the parsing, most likely the generic patterns would end up picking up agents like the one above (it doesn't happen in this case, because the Generic file for Firefox doesn't have "Win98" in it, but that should probably be added), so they would still be parsed as Firefox on whichever platform, but they wouldn't have a Firefox version number.

Or, we could try to treat these as invalid patterns (like how browscap 6022 treats that fake pattern above), but that may be more difficult because that generic pattern file wouldn't be able to pick up these agents (which may mean that most patterns would have to be removed from it, or we'd at least have to be selective about which platforms show up in that file (like no Win98 for example, since we know that only certain version ranges of Firefox support Win98)).

If we choose not to allow these platforms in version files where they shouldn't be, what should be done with the Kernel extension projects?

IMO, I think the best approach may be to drop the platforms from the files for the versions that Firefox officially stopped supporting that platform, and make sure the generic file has all of the platforms in it so that the UAs still parse as Firefox, but will be version-less. (this is the approach I've been taking with recent changes I've been making).

Thoughts?  What's the official stance here?

jaydiablo:
This fixes #1381.

There’s a lot going on here, I spent a lot of time looking into version history of firefox across different platforms and grabbed a lot of user-agents from our logs to try and add as much coverage to the firefox patterns as I could.

I’ll address the specific changes with inline comments.

I did run these changes vs. master against a number of large lists of useragents (40k+ from parser test suites, 50k that I collected from our logs, 1k for each previous month of 2017 from our logs) and confirmed the differences were expected.  For the most part, parsing improves for all useragents. There is a loss of a version number for a handful of agents, but that was intentional, as I’ll describe in the comments.

I also added a fixture file that contains “invalid versions”, which I mention in #1381.  I feel it’s a good way to prevent regression of adding back platforms to specific versions of a browser that shouldn’t be added for whatever reason.  I put comments in that file for every test case to describe why it was added, with a source.  I only included a handful of properties for this, since they’re really only testing that the version of the browser is an expected value (0.0 in most cases) or that the “Generic” pattern was used.  I feel that this may help with maintenance of the tests in the future (not having so many properties to modify if they change).


resources/user-agents/browsers/firefox/firefox-3-1.json
```diff
@@ -23,7 +23,8 @@
         "Tables": "true",
         "Cookies": "true",
         "JavaScript": "true",
-        "CssVersion": "3"
+        "CssVersion": "3",
+        "Beta": "true"
```
jaydiablo: 3.1 never had an official release (3.5 was the next major release), so I marked this as beta.

https://en.wikipedia.org/wiki/Firefox_3.5#Beta_and_release

resources/user-agents/browsers/firefox/firefox-3-6.json
```diff
@@ -159,13 +164,6 @@
           ]
         },
         {
-          "match": "Mozilla/5.0 (#PLATFORM#; rv:1.9.2*) Gecko* Firefox *",
```
jaydiablo: I moved this to the generic file since it doesn't have #MAJORVER# or #MINORVER# tokens.

resources/user-agents/browsers/firefox/firefox-3-7.json
```diff
@@ -23,7 +23,7 @@
         "Cookies": "true",
         "JavaScript": "true",
         "CssVersion": "3",
-        "Beta": "true"
+        "Alpha": "true"
```
jaydiablo: I switched this one to Alpha, since everything I could find showed that 3.7 was an alpha release before they decided to change the version to 4.0.

https://en.wikipedia.org/wiki/Firefox_version_history#Release_history (Firefox 4.0 section of that table).

resources/user-agents/browsers/firefox/firefox-3-8.json
```diff
@@ -1,109 +0,0 @@
-{
```
jaydiablo: As far as I can tell, there was never a 3.8 release tagged (as alpha or otherwise). In our logs I only found 2 useragents with a Firefox 3.8 version number (this is since 2009 or so).

Firefox's version was bumped to 4.0 during the 3.7 alpha release cycle, so I don't see why 3.8 would have ever been tagged.

resources/user-agents/browsers/firefox/firefox-4-2.json
```diff
@@ -3,7 +3,7 @@
   "versions": ["4.2"],
   "sortIndex": 3163,
   "lite": false,
-  "standard": true,
+  "standard": false,
```
jaydiablo: I removed this from Standard since it's an Alpha release.

It doesn't show up on this table: https://en.wikipedia.org/wiki/Firefox_version_history#Release_history

But I did find some discussion about people downloading it from the nightlies.  5.0 was the next official release after 4.0.

resources/user-agents/browsers/firefox/firefox-49-52.json
```diff
@@ -30,7 +30,7 @@
           "match": "Mozilla/5.0 (#PLATFORM#) Gecko* Firefox/#MAJORVER#.#MINORVER#*",
           "device": "Linux Desktop",
           "platforms": [
-            "Ubuntu_64", "Ubuntu_12_04", "Ubuntu_32", "Ubuntu",
```
jaydiablo: This removed Ubuntu platform won't actually catch any useragents because the platform's pattern is defined as `*Ubuntu/12.04*` which doesn't seem to ever happen in this part of the UA (it does happen after the `Gecko` keyword in older versions of firefox.

resources/user-agents/browsers/firefox/firefox-generic.json
```diff
@@ -26,77 +26,86 @@
...
           "platforms": [
-            "FreeBSD_64", "FreeBSD", "HPUX", "IRIX64", "OpenBSD_64_2", "OpenBSD", "SunOS", "OS2"
+            "FreeBSD_64", "FreeBSD",
+            "OpenBSD_64", "OpenBSD",
+            "NetBSD", "DragonFlyBSD",
+            "HPUX", "IRIX64", "BeOS", "SunOS", "OS2"
           ]
         },
         {
           "match": "Mozilla/4.0 (#PLATFORM#) Gecko* Firefox/*",
```
jaydiablo: I should mention that I'm suspicious of these `Mozilla/4.0` patterns, especially on newer versions of Firefox.  I don't think Firefox uses this prefix (not sure if it ever really did).

We do see a handful of useragents with this prefix that browscap has detected as Firefox, across a lot of different versions of Firefox (but at a much lower frequency than the `Mozilla/5.0` prefix) so I kept the patterns here (but didn't add any tests for it).

Perhaps re-visiting the Firefox source code to see if `Mozilla/4.0` could every happen legitimately is in order before making a decision on these.

resources/user-agents/browsers/firefox/firefox-generic.json
```diff
@@ -168,82 +170,28 @@
...
+            "WinXPa_32", "Vista_32", "Win7_32", "Win8_32"
+          ],
           "properties": {
-            "isAnonymized": "true"
-          },
-          "device": "general Tablet",
-          "platforms": [
-            "AndroidTablet"
-          ]
+            "Version": "3.6",
```
jaydiablo: These are those patterns that I moved from the 3.5 and 3.6 files since they had no version tokens.

resources/user-agents/browsers/misc/alienforce.json
```diff
@@ -0,0 +1,39 @@
+{
```
jaydiablo: For all of these new browsers I didn't add any versions to keep the pattern count low. Could be added later if necessary.

resources/user-agents/browsers/palemoon/palemoon-2-x.json
```diff
@@ -1,18 +1,18 @@
 {
-  "division": "Palemoon #MAJORVER#.#MINORVER#",
+  "division": "PaleMoon #MAJORVER#.#MINORVER#",
```
jaydiablo: It was confusing that the comment used a lower case "m" and the Browser used an uppercase "M" here, so I changed the comment and division headers to match the "Browser" property.

tests/fixtures/issues/issue-000-invalid-versions.php
```diff
@@ -0,0 +1,604 @@
+<?php
+
+return [
```
jaydiablo: This is the invalid-versions fixture I've mentioned. Each entry is either from our logs, or taken from exisiting tests (and moved here). I tried to make sure each entry has a comment about why it's here, with a link to a source.

tests/fixtures/issues/issue-033.php
```diff
@@ -15,7 +15,7 @@
             'MinorVer' => '6',
             'Platform' => 'MacOSX',
             'Platform_Version' => '10.5',
-            'Platform_Description' => 'Mac OS X',
+            'Platform_Description' => 'Mac OS X for Power PC',
```
jaydiablo: With the addition of the `OSX_PPC` platform, there are a lot of tests with this change.

tests/fixtures/issues/issue-1360.php
```diff
@@ -1,278 +1,13 @@
 <?php
 
 return [
-    'issue-1360-A' => [
-        'ua' => 'Mozilla/5.0 (Windows; U; WinNT4.0; bg; rv:1.8.1.4) Gecko/20100401 Firefox/4.0',
```
jaydiablo: A number of agents here (that I added earlier, heh) that are bad platform/browser version combinations and have been moved to the invalid-versions fixture.

tests/fixtures/issues/issue-1360.php
```diff
@@ -1,278 +1,13 @@
...
-        'standard' => false,
-    ],
-    'issue-1360-F' => [
         'ua' => 'Mozilla/5.0 (X11; NetBSD amd64; rv:30.0) Gecko/20100101 Firefox/30.0',
         'properties' => [
             'Comment' => 'Firefox 30.0',
             'Browser' => 'Firefox',
             'Browser_Type' => 'Browser',
-            'Browser_Bits' => '32',
+            'Browser_Bits' => '64',
```
jaydiablo: I overlooked the "amd64" part of these useragents at the time.

tests/fixtures/issues/issue-267.php
```diff
@@ -3181,59 +3181,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-267-BI' => [
-        'ua' => 'Mozilla/6.0 (Windows NT 6.2; WOW64; rv:16.0.1) Gecko/20121011 Firefox/16.0.1',
```
jaydiablo: `Mozilla/6.0` prefix, moved to "invalids" fixture.

tests/fixtures/issues/issue-511.php
```diff
@@ -2495,15 +2495,15 @@
     'issue-511-BE' => [
         'ua' => 'Mozilla/5.0 (Android; Mobile; rv:33.1) Gecko/33.1 Firefox/33.1',
         'properties' => [
-            'Comment' => 'Firefox 33.0',
```
jaydiablo: Since I moved Firefox Android to its own file, I was able to add these ".1" releases (seems there were only two of them so far): https://en.wikipedia.org/wiki/Firefox_for_Android#Release_history

Here's the release announcement for this version: https://www.mozilla.org/en-US/firefox/android/33.1/releasenotes/

tests/fixtures/issues/issue-618.php
```diff
@@ -4930,59 +4930,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-618-CP' => [
-        'ua' => 'Mozilla/5.0 (Windows NT 6.2; WOW64; rv:29.0) Gecko/20100101 /',
```
jaydiablo: Moved to "invalids" fixture since it lacks "Firefox".

tests/fixtures/issues/issue-814.php
```diff
@@ -3499,59 +3499,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-814-BQ' => [
-        'ua' => 'Mozilla/5.0 (Windows NT 10.0; WOW64; rv:41.0) Gecko/20100101 Firefox/AA5D',
```
jaydiablo: This is one of those useragents with the alphanumeric Firefox version (a few more follow this one). These were moved to the "invalid-versions" fixture.

tests/fixtures/issues/issue-821.php
```diff
@@ -478,59 +478,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-821-Y' => [
-        'ua' => 'Mozilla/5.0 (Windows NT 10.0; WOW64; rv:41.0) Gecko/20100101 Firefox/4DF0',
```
jaydiablo: More alpha-numeric Firefox version.

tests/fixtures/issues/issue-865.php
```diff
@@ -2545,59 +2545,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-865-AW' => [
-        'ua' => 'Mozilla/5.0 (Windows NT 6.3; rv:36.0) Gecko/20100101 Firefox',
```
jaydiablo: Ultimately I ended up removing this one from the test suite. I played with changing the Generic pattern to not have the slash after "Firefox" (and just be `Firefox*`) but this caused a lot of other issues since it shortened the generic pattern too much. I could have added a pattern just for this useragent so that it would still detect as Firefox (no version), but it didn't seem worth it.

tests/fixtures/issues/issue-900.php
```diff
@@ -15848,112 +15848,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-900-LC' => [
-        'ua' => 'Mozilla/6.0 (Macintosh; I; Intel Mac OS X 11_7_9; de-LI; rv:1.9b4) Gecko/2012010317 Firefox/10.0a4',
```
jaydiablo: I believe this one ended up as an "invalid-version" since there are a number of issues with it (bad OSX version (currently), `rv` and `Firefox` versions don't match, `Gecko` tail isn't correct).

tests/fixtures/issues/issue-900.php
```diff
@@ -15848,112 +15848,6 @@
...
-            'Device_Brand_Name' => 'Apple',
-            'RenderingEngine_Name' => 'Gecko',
-            'RenderingEngine_Version' => '10.0',
-            'RenderingEngine_Maker' => 'Mozilla Foundation',
-        ],
-        'lite' => false,
-        'standard' => true,
-    ],
-    'issue-900-LD' => [
-        'ua' => 'mozilla/3.0 (Windows NT 6.1; rv:2.0.1) Gecko/20100101 Firefox/5.0.1',
```
jaydiablo: Here's that rare `mozilla/3.0` prefixed UA.  I believe this is a fake.

I removed it from the suite because right now it detects as Netscape (Netscape is one of the few pattern files that has a `Mozilla/3.0` prefix, but is also probably the only legitimate one).

tests/fixtures/issues/issue-900.php
```diff
@@ -16007,59 +15901,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-900-LF' => [
-        'ua' => 'Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.9.2.3) Gecko/20100401 Mozilla/5.0 (X11; U; Linux i686; it-IT; rv:1.9.0.2) Gecko/2008092313 Ubuntu/9.25 (jaunty) Firefox/3.8',
```
jaydiablo: We saw this in our logs too, but it's one of two `3.8` patterns. I believe this was moved to "invalid-versions" file.

PR #1426, opened 2017-05-19
Improving UC Browser Detection with coverage

### LINKED ISSUE #1425: Organize/Consolidate UC Browser patterns
When I was adding some mis-detections for #1424 I noticed that there were a number of patterns in the UC Browser files that seemed redundant and/or could be grouped together better so that it's easier to add patterns to the files.

Before doing any sort of consolidation, I wanted to make sure that I wasn't breaking detection so wanted to add some tests for missing coverage.  Wanted an issue number for those tests.

jaydiablo:
Fixes #1425.

This adds a number of patterns/platforms to improve UC Browser detection. It does remove some unnecessary patterns that weren’t matching anything (Android_C and Android_D patterns doubled up in some places mainly).

I also tried to standardize the UC Browser patterns as much as possible across all files without impacting tests at all.

I added a UC Browser 6 file that just covers their Windows version of the browser (which I downloaded to a windows machine and confirmed the UA for).

I added some missing patterns to the generic file and also added a generic file for the UC Mini browser to catch any versions that aren’t defined in the existing file.

I collected about 130,000 different UC Browser agents that I was able to extract from our logs which is what I used to determine if/how these patterns could be improved.

I only added tests for patterns that weren’t covered already, and for which browscap has a device defined (no “general mobile phone” tests if the UA contained device information at all).

I did benchmark these changes against `master` and there was no significant difference between these changes and `master` (sometimes this branch out performed master, sometimes master was faster depending on the UAs in the flies I tested against).

Here’s the output of just the browser name comparison of that large list of UAs that I mentioned:

```
+------------------------+---------------------+----------------------------+
|  Expected Browser Name | browscap-php-3-full | browscap-php-3-full-master |
+------------------------+---------------------+----------------------------+
| ucbrowser (135447)     | ucbrowser (135447)  | ucbrowser (135000)         |
|                        |                     | android (359)              |
|                        |                     | [no value] (43)            |
|                        |                     | chrome (42)                |
|                        |                     | androidwebview (2)         |
|                        |                     | mobilesafariuiwebview (1)  |
+------------------------+---------------------+----------------------------+
| [no value] (94)        | [no value] (94)     | [no value] (94)            |
|                        |                     |                            |
+------------------------+---------------------+----------------------------+
| kindlefire (3)         | kindlefire (3)      | kindlefire (3)             |
|                        |                     |                            |
+------------------------+---------------------+----------------------------+
```


tests/fixtures/issues/issue-317.php
```diff
@@ -1657,7 +1657,7 @@
             'MajorVer' => '9',
             'MinorVer' => '4',
             'Platform' => 'Android',
-            'Platform_Version' => '4.0',
```
jaydiablo: The changes in these test files were because the UA that these are testing are of this variety:

>UCWEB/2.0 (Linux; U; Opera Mini/* #DEVICE#)

Where there is no actual platform indicator. It seems that the android version used was based on the phone's capabilities.  Many of the devices that were detected this way support multiple versions of android, so I moved those to just a generic Android platform (which is why this changed to `unknown`). In other cases, the source I had for the phone showed a different version than what was defined.

In general it's probably better to move all of these to a generic "Android" platform with no version, since there's really no way to tell which version of android is installed on the phone from just the UA.

I can make this change if that's the way we want to go with that.

resources/user-agents/browsers/uc-browser/uc-web-2.json
```diff
@@ -56,58 +62,119 @@
...
+          },
           "platforms": [
-            "Android_E_4_4", "Android_E_4_3", "Android_E_4_2", "Android_E_4_1", "Android_E_4_0",
-            "Android_E_2_3", "Android_E_2_2", "Android_E_2_1",
-            "Android_E"
+            "Android_4_2"
           ]
         },
         {
-          "match": "UCWEB/*(Linux; U; Opera Mini/*) U2/* UCBrowser/#MAJORVER#.#MINORVER#*",
```
jaydiablo: This is where most of those Opera Mini patterns are defined that I mention below.

PR #1431, opened 2017-05-21
Removing iOS < 3.2 for iPad

jaydiablo:
For some reason I was thinking about this today.

The original iPad released with iOS 3.2 (https://en.wikipedia.org/wiki/IPad_(1st_generation)#Software) and now that iphone/ipad are pretty much separated in browscap files, I was curious if there were any instances of iOS < 3.2 + iPad in the patterns.

I found a few, so removed them.

I also removed a few of these iOS_C platforms that aren't used in resource files at all anymore.

No tests covered these removals (they shouldn't, since iPad + iOS < 3.2 shouldn't exist).

### DIFF
--- a/resources/platforms.json
+++ b/resources/platforms.json
@@ -653,33 +653,6 @@
       "standard": true,
       "inherits": "iOS"
     },
-    "iOS_C_2_0": {
-      "match": "*CPU OS 2* like Mac OS X*",
-      "lite": false,
-      "standard": false,
-      "inherits": "iOS",
-      "properties": {
-        "Platform_Version": "2.0"
-      }
-    },
-    "iOS_C_2_1": {
-      "match": "*CPU OS 2_1* like Mac OS X*",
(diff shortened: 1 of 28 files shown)


jaydiablo:
Reading about this some more, and there are some others that can probably be removed, I'll remove and run tests to confirm.

jaydiablo:
So, according to the wikipedia article about iOS version history (https://en.wikipedia.org/wiki/IOS_version_history) there were a few other iOS version/device combinations that aren't possible.

iOS 4.0 and 4.1 were iPod/iPhone only releases (the first 4.x version for the iPad was 4.2): https://en.wikipedia.org/wiki/IOS_version_history#iOS_4

iOS 3.2 (which the iPad released with) was an iPad only release: https://en.wikipedia.org/wiki/IOS_version_history#iPhone_OS_3

I've removed these combinations where possible, no tests covered these removals.

PR #1435, opened 2017-05-30
Opera Update (desktop/mobile/mini)

### LINKED ISSUE #1434: Opera Detection/Coverage
Opera was next on my list to look at for device consolidation, I wanted to add some coverage as part of the process, which revealed some detection issues.

Will attach the new tests to this issue.

jaydiablo:
Fixes #1434.

As mentioned in the ticket, I wanted to add coverage to Opera patterns as part of a device consolidation.

Like other large updates I’ve done, I pulled down as many Opera related useragents from our logs as I could (I also used useragents from other parsers’ test suites).

Also, like the work I’ve done with Safari and Firefox, I’m removing some patterns that weren’t possible to have coverage for, mainly versions + platform combinations that shouldn’t be possible based on what versions of different platforms Opera has supported over time.

I used Opera’s developer blog, their public archive and wikipedia to determine what versions of Opera supported which platforms and when (I did this for the desktop browser, mobile browser and the Opera Mini browser since there was a lot of cross-contamination between the three browsers).

Overall I believe the number of patterns in the files reduced (the full version of the cache files reduced by about 2 megabytes compared to `master`).  Benchmarks look to be close to the same as `master` but are a tiny bit slower in some cases (10ths of a second for ~1,000 useragents).

I also compared the processing of `master` and this branch against ~150,000 unique useragents that I collected during this process to be sure that these changes don’t negatively affect parsing. Other than removed version numbers (based on research I did) this branch does parse these agents better:

```
+------------------------+----------------------+----------------------------+
|  Expected Browser Name | browscap-php-3-full  | browscap-php-3-full-master |
+------------------------+----------------------+----------------------------+
| operamini (86328)      | operamini (86328)    | operamini (86320)          |
|                        |                      | mobilesafariuiwebview (7)  |
|                        |                      | [no value] (1)             |
+------------------------+----------------------+----------------------------+
| opera (67883)          | opera (67883)        | opera (66389)              |
|                        |                      | [no value] (1316)          |
|                        |                      | chrome (132)               |
|                        |                      | ie (42)                    |
|                        |                      | firefox (2)                |
|                        |                      | netscape (1)               |
|                        |                      | blackberry (1)             |
+------------------------+----------------------+----------------------------+
| ucbrowser (67)         | ucbrowser (67)       | operamini (60)             |
|                        |                      | generalcrawlers (4)        |
|                        |                      | opera (3)                  |
+------------------------+----------------------+----------------------------+
| googleappengine (37)   | googleappengine (37) | googleappengine (35)       |
|                        |                      | opera (2)                  |
|                        |                      |                            |
+------------------------+----------------------+----------------------------+
| [no value] (35)        | [no value] (35)      | [no value] (35)            |
|                        |                      |                            |
+------------------------+----------------------+----------------------------+
| nintendo (10)          | nintendo (10)        | nintendo (10)              |
|                        |                      |                            |
+------------------------+----------------------+----------------------------+
| ie (6)                 | ie (6)               | ie (6)                     |
|                        |                      |                            |
+------------------------+----------------------+----------------------------+
| firefox (3)            | firefox (3)          | opera (3)                  |
|                        |                      |                            |
|                        |                      |                            |
+------------------------+----------------------+----------------------------+
| nutch (1)              | nutch (1)            | nutch (1)                  |
|                        |                      |                            |
+------------------------+----------------------+----------------------------+
| toutiaospider (1)      | toutiaospider (1)    | toutiaospider (1)          |
|                        |                      |                            |
+------------------------+----------------------+----------------------------+
```

That Firefox change (from Opera) is the result of agents like this:

>Mozilla/5.0 (Windows; U; Windows NT 6.1; en-US; rv:1.9.2.13) Gecko/20101213 Opera/9.80 (Windows NT 6.1; U; en-us) Presto/2.7.62 Version/11.01
>Mozilla/5.0 (Windows; U; Windows NT 6.1; en-US; rv:1.9.2.13) Gecko/20101213 Opera/9.80 (Windows NT 6.1; U; zh-tw) Presto/2.7.62 Version/11.01
>Mozilla/5.0 (Windows; U; Windows NT 6.1; en-US; rv:1.9.2.13) Gecko/20110111 Opera/9.80 (Windows NT 6.1; U; zh-tw) Presto/2.7.62 Version/11.01

Which I removed the pattern for since they’re a weird combo-agent (platform appears twice).

The tests that show as removed were almost all moved to the `invalid-versions` fixture file with a description as to why the agent should be parsed as `Opera Generic` instead of a proper version (and links to sources where possible).

A few other changes in the tests were because I added patterns that a re now catching some that were `Default Browser` previously, or because of some minor data changes (like removing the `Beta` flag for Opera Mini 8+).

I should mention that I removed the `Opera 9.80` file entirely, as it doesn’t appear that Opera ever released that version, and when I had it in there, the only patterns in the file that were covered by my ~150,000 useragents were the ones that started with `Opera/9.80`, which is a prefix that later versions of Opera use, but it would fall back to `9.80` if the version later in the agent didn’t match anything applicable (or was missing).

I did make sure that the tests I’ve added are only for desktop, specific mobile devices, or generic mobile devices if the device information isn’t present in the UA.


resources/user-agents/browsers/opera-mini/opera-mini-10-0-on.json
```diff
@@ -0,0 +1,39 @@
+{
+  "division": "Opera Mini #MAJORVER#.#MINORVER#",
+  "versions": [28, 27, 26, 25, 24, 23, 22, 21, "20.1", 20, 19, 18, 17, 16, 15, 14, 13, 12, 11, "10.2", "10.1", 10],
+  "sortIndex": 1931,
+  "lite": false,
+  "standard": true,
+  "userAgents": [
+    {
+      "userAgent": "Opera Mini #MAJORVER#.#MINORVER#",
+      "engine": "Presto_2_8",
```
mimmi20: Is the engine still Presto?

resources/user-agents/browsers/opera/opera-11-00.json
```diff
@@ -1,6 +1,6 @@
 {
   "division": "Opera #MAJORVER#.#MINORVER#",
-  "versions": ["11.01", "11.00"],
+  "versions": ["11.01", "11.1", "11.00", 11],
```
mimmi20: version `11.1` is added in the `opera-11-10` file too. Please remove it here.

resources/user-agents/browsers/opera/opera-12-10-to-12-50.json
```diff
@@ -0,0 +1,205 @@
+{
+  "division": "Opera #MAJORVER#.#MINORVER#",
+  "versions": ["12.18", "12.17", "12.16", "12.15", "12.14", "12.13", "12.12", "12.11", "12.10"],
```
mimmi20: the file name does not match to the versions.

resources/user-agents/browsers/opera-mini/opera-mini-10-0-on.json
jaydiablo: According to wikipedia, Opera Mini still uses Presto as of March 2015: https://en.wikipedia.org/wiki/Opera_Mini#Standards_support

However, at the same time it says it uses a mix of Blink and Presto (side bar): https://en.wikipedia.org/wiki/Opera_Mini

So, hard to say, typically Opera Mini can/will be server rendered using Opera's proxy servers.  Sounds like it uses Presto in that case, but maybe uses Blink for on-device rendering?

mimmi20: Then we could let Presto stay as it is.

PR #1439, opened 2017-05-31
Consolidating devices and adding coverage for blackberry browser

### LINKED ISSUE #1438: Blackberry device consolidation/coverage
Have some tests to add for Blackberry browser.

jaydiablo:
Fixes #1438.

Blackberry had a number of devices to consolidate.

Added coverage by finding as many blackberry UAs as I could from our logs.

I also took the time to update the names/pointing methods for any devices that showed up in the tests I added, this required updating a few other tests that the devices appeared on.

I did end up adding some new patterns to cover some of the missing devices in the agents I found. However, I did also remove some what looked like incomplete patterns from the 6.x and 7.x files. None of the agents I found hit those patterns, so not really sure why they were there.


resources/user-agents/browsers/blackberry/blackberry-7-1.json
```diff
@@ -28,108 +28,52 @@
       },
       "children": [
         {
-          "match": "Blackberry9800/#MAJORVER#.#MINORVER#*",
-          "device": "Blackberry 9800"
-        },
-        {
-          "match": "Mozilla/5.0 (BlackBerry; ?; BlackBerry 9105*applewebkit* (*khtml*like*gecko*) Version/#MAJORVER#.#MINORVER#*Safari/*",
```
jaydiablo: These are the "incomplete" patterns I was talking about.  They're duplicates of patterns below, they just seem to be lacking a few characters between the device and `*applewebkit`. After running browscap against the useragents I found, and looking at the coverage report, these were never hit (the longer pattern below was always used). Since there were no tests for these patterns, and I couldn't find any UAs that hit them, I figured it was safe to remove.

resources/devices/mobile-phone/rim.json
```diff
@@ -259,14 +259,27 @@
...
         }
     },
+    "Blackberry 8130m": {
+        "standard": false,
+        "properties": {
+            "Device_Type": "Mobile Phone",
+            "Device_Name": "Pearl 8130m",
+            "Device_Code_Name": "BlackBerry 8130m",
+            "Device_Maker": "Research In Motion Limited",
+            "Device_Pointing_Method": "unknown",
```
mimmi20: Does this device also has a `clickwheel`?

jaydiablo: Ah, it should, I'll update.

PR #1441, opened 2017-05-31
Device consolidation and coverage for Dolfin browser

### LINKED ISSUE #1440: Samsung Dolfin device consolidation + coverage
Have some tests to add for this browser.

jaydiablo:
Fixes #1440.

Mostly just device consolidation with coverage.  I did add and cleanup a few devices, but didn’t go crazy with it. :)


resources/user-agents/browsers/dolfin/dolfin-1-0-1-5.json
```diff
@@ -28,108 +28,45 @@
...
-          "device": "Samsung SGH-I900"
+          "match": "#DEVICE#/*Dolfin/#MAJORVER#.#MINORVER#*MIDP*",
+          "devices": {
+            "SAMSUNG-GT-B7722": "Samsung GT-B7722",
+            "SAMSUNG-GT-S5620": "Samsung GT-S5620",
+            "SAMSUNG-SGH-T528g": "Samsung SGH-T528g"
+          }
+        },
+        {
+          "match": "#DEVICE#*?Dolphin/#MAJORVER#.#MINORVER#*",
```
jaydiablo: I couldn't find any useragents that covered these patterns (or any that started with "Samsung-" and contained "Dolphin". 

Can remove if we want, I left them in just in case, and they don't really get looked at by the parser much since they start with "Samsung".

PR #1454, opened 2017-06-02
Fixing some mis-detections for Iron browser, separating "MobileIron" and "Iron Mobile"

### LINKED ISSUE #1453: "Iron Mobile" and "MobileIron" are actually two different things
While looking at UAs for #1448 I noticed a number of "Iron" browsers that were being detected as Chrome.  Aside from the Desktop ones (which I'm planning to fix) there were a few mobile ones as well.

There's these patterns: https://github.com/browscap/browscap/tree/master/resources/user-agents/browsers/iron-mobile

But they're looking for "MobileIron" in the UA. However, I believe "MobileIron" isn't an SRWare product, but instead is this company: https://www.mobileiron.com

SRWare does have an Android browser: https://play.google.com/store/apps/details?id=org.iron.srware&hl=en but no iOS flavor as far as I can tell.

MobileIron has both Android and iOS offerings.

I have a PR coming to address this, as well as improve coverage for the Iron patterns.

jaydiablo:
Fixes #1453.

Fixed some cases where these Iron UAs were being detected as Chrome (by adding some platforms, versions and a slightly modified pattern).

I did combine some patterns in these Iron files into a single pattern that works for both variations which should reduce the pattern count some.  There are some patterns in the higher versions of Iron that don’t seem to ever get touched by actual UAs, but I don’t have access to enough Iron UAs to make the decision to remove them entirely.

This also moves the “MobileIron” patterns to a different “maker” and adds some patterns to catch the “Iron Mobile” Android browser, as discussed in #1453.


resources/user-agents/browsers/iron-mobile/iron-mobile-generic.json
```diff
@@ -24,69 +24,16 @@
       },
       "children": [
         {
-          "match": "Mozilla/5.0 (#PLATFORM#) applewebkit* (*khtml*like*gecko*) Chrome/* MobileIron/* Mobile Safari/*",
+          "match": "Mozilla/5.0 (#PLATFORM#) applewebkit* (*khtml*like*gecko*) Chrome/* Iron* Safari/*",
           "platforms": [
             "Android_7_1", "Android_7_0",
-            "Android_6_1", "Android_6_0",
```
jaydiablo: Apparently there was no Android 6.1 version: https://en.wikipedia.org/wiki/Android_version_history

So I've been starting to remove it when I encounter it.  Should probably make a PR to remove across all files at some point.

resources/user-agents/browsers/iron/iron-45-49.json
```diff
@@ -28,26 +28,29 @@
       "children": [
         {
           "match": "Mozilla/5.0 (#PLATFORM#) applewebkit* (*khtml*like*gecko*)*Iron/#MAJORVER#.*Safari/*",
+          "engine": "WebKit",
```
jaydiablo: These (and the next block of 3 that follow these) are the patterns that I mention that I don't think are used by later versions of Iron.

I changed this one to WebKit since there's no Chrome in the pattern, but I don't think it matters for these versions.

I'd vote to remove them from this file, any later versions, and the 40-44 file as well.

The coverage report better reveals which are touched and by what versions.

PR #1500, opened 2017-06-08
Cleaning up Iron patterns some more (just desktop), adding coverage

### LINKED ISSUE #1499: Remove Iron patterns, increase coverage
I ran into a few more Iron UAs that were being detected as Chrome.

I did address this recently in #1454, but I didn't really spend any time looking at the coverage, or searching through more versions of Chrome.

I pulled a bunch of UAs from our logs that have previously detected as Iron, as well as any I could find in all version of Chrome that were detected as Chrome and ran them through browscap.

Then I spent some time looking at what should be valid and not based on archive.org snapshots of the SRWare website/forums.  I was able to remove a number of platforms that aren't supported (like OSX < 10.5 or Windows < XP).

Then I also used the coverage report after generating a bunch of tests to see which sets of patterns are touched by Iron useragents in the wild.  These Iron files have a lot of copy/pasted patterns in them from older versions that don't seem to apply to newer versions of the Iron agents, so a lot can be removed here.

jaydiablo:
Fixes #1499

Removed a bunch of unused patterns after analyzing ~1500 different Iron UAs from our logs.

Overall, this improves detection compared to `master` as I did add some new platforms.

`master` sizes:

2,367,990 lines
214,169 patterns

This branch sizes:

2,338,359 lines (-29,631)
211,931 patterns (-2,238)


tests/fixtures/issues/issue-900.php
```diff
@@ -22155,59 +22155,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-900-TL' => [
```
jaydiablo: I removed this test because it was getting picked up by the Generic file because I added the "Windows" platform.  I didn't think it was worth modifying the test since there's not really a windows version in this UA.

tests/fixtures/issues/issue-1453.php
```diff
@@ -1432,59 +1432,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-1453-AB' => [
```
jaydiablo: This test got moved to the invalid-versions fixture since it's Windows 2000, which as far as I can tell, was never supported by Iron (see the invalid versions fixture for more info).

tests/fixtures/issues/issue-1453.php
```diff
@@ -4042,7 +3989,7 @@
             'MajorVer' => '31',
             'MinorVer' => '0',
             'Platform' => 'MacOSX',
-            'Platform_Version' => '10',
+            'Platform_Version' => '10.11',
```
jaydiablo: I missed this when I made this test. I look over the tests looking for an "unknown" platform version where there shouldn't be one, but I forgot that OSX typically gets caught as version "10" due to the "OSX" platform.

PR #1543, opened 2017-08-08
Build stages

mimmi20:
I would like to use the travis build stages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: php
 sudo: false
 
 php:
-  - 7.0
   - 7.1
   - nightly
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,48 +6,53 @@ php:
   - 7.1
   - nightly
 
-os:
-  - linux
+os: linux
 
 env:
   - COMPOSER_FLAGS="--prefer-lowest"
   - COMPOSER_FLAGS=""
-
-matrix:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,53 +1,58 @@
 language: php
 sudo: false
+dist: trusty
```
asgrim: This is implied anyway isn't it now? I thought Travis had already migrated default to `trusty`?

.travis.yml
```diff
@@ -1,53 +1,58 @@
...
 
 php:
   - 7.0
```
asgrim: Don't think we need to test on 7.0 any more. Let's bump min to 7.1.

.travis.yml
mimmi20: According to this post (https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming?utm_source=web&utm_medium=banner&&utm_campaign=trusty-default) the migration shall be finished at 1st of september.

.travis.yml
mimmi20: removed

.travis.yml
mimmi20: removed

mimmi20:
@asgrim Could you review again?

asgrim:
Thanks!

PR #1556, opened 2017-08-14
Adding some 2345 patterns

### LINKED ISSUE #1555: 2345 Explorer mis-detections + tests
Also part of #1551

jaydiablo:
Fixes #1555.

This adds some patterns and versions for 2345 Explorer, as well as adds 2345 Chrome (which seems to be an earlier version of 2345 Explorer).

I grabbed as many UAs from our logs as I could find to help deduce what the patterns should be at various versions, and then added as many of those as necessary as tests to cover the patterns.


jaydiablo:
I'm not sure why the Travis test segfaulted:

https://travis-ci.org/browscap/browscap/jobs/264509925#L655

mimmi20 (approved):

PR #1568, opened 2017-09-02
remove grep and diff commands which are not working

### LINKED ISSUE #1567: GrepCommand, DiffCommand
These two commands are actually not working. Browscap is called there in the V1 syntax.

Does anyone use these commands? If not I would prefer to remove them.

mimmi20:
I removed these two commands.

Fixes #1567.


asgrim:
LGTM :+1:

PR #1570, opened 2017-09-05
New Chedot version/pattern

### LINKED ISSUE #1569: Missing Chedot pattern
Was looking at mis-detections from August and noticed a new version of Chedot browser that is being caught as Chrome:

>Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.78 Chedot/8.0.0 Safari/537.36

I downloaded and confirmed this UA.

I'll add this pattern, as well as some tests from other Chedot UAs we have in our logs.

jaydiablo:
Fixes #1569.

Added a new pattern for Chedot 8.0 which varies slightly from the existing patterns.

Added a 3.4 file as well, to catch some UAs we’ve seen. I didn’t think it was good to add 3.4 to the file that has version 43 in it, as we’ve never seen the `Chrome/3.4` style pattern for this version of Chedot.

Also renamed the `28-on` file since it only contains one version (43.0), which is also the only version in that style of UA that we’ve seen.

Added tests from UAs in our logs to cover any uncovered patterns.


resources/user-agents/browsers/chedot/chedot-43-0.json
```diff
@@ -47,7 +49,9 @@
             "Win10_x64", "Win10_64", "Win10_32",
             "Win8_1_x64", "Win8_1_64", "Win8_1_32",
             "Win8_x64", "Win8_64", "Win8_32",
-            "Win7_x64", "Win7_64", "Win7_32"
+            "Win7_x64", "Win7_64", "Win7_32",
+            "Vista_64", "Vista_32",
```
jaydiablo: Ah, and I did add Win Vista and XP to these patterns because their site mentions that the browser is available for these platforms: http://landing.chedot.com

Issue #961, opened 2016-05-01
Regression between version 6013 and 6014


wggdeveloper:
Using all the UAs from useragentstring.com I compiled a [list](https://github.com/browscap/browscap/files/244120/regression.txt) of 616 (5.72%) strings which was detected in 6013 but not in 6014. The second column contains the matching patterns from 6013.

wggdeveloper:
Here is another [list](https://github.com/browscap/browscap/files/244151/regression.txt) compiled from the tests of ua-parser/uap-core. There could be a couple of specifically crafted test strings, but most of them are real UAs.

mimmi20:
@wggdeveloper These Uas are not detected, because we removed these rules. There were false positives and wrong detected fakes in there.

Your lists are too long and include too much duplicate useragents. Please remove the following from your lists:
- user agents with changes in Gecko release dates only
- user agants with changes in the micro version or alpha/beta flags only
- user agents with changes in the language key only
- user agants with different size of spaces inside only
- every user agent who looks like a fake
- every user agant with not existing versions, Windows NT codes, combinations of browser/os
- additional information like Linux system (e.g. red hat, fedora, debian), other software (mega upload)

For the remaining Uas please add a suggestion what the browser name should be.

jaydiablo:
This is an old issue, but I was curious how the most recent version (dev-master as of this writing) of browscap handles the agents in the first file.  Almost all of them parse as "Default Browser", except this batch:

>Mozilla/5.0 (Macintosh; ; Intel Mac OS X; fr; rv:1.8.1.1) Gecko/20061204 Opera
Mozilla/5.0 (Macintosh; Intel Mac OS X; U; nb; rv:1.7.5) Gecko/20041110
Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10.5; en-au; rv:1.9.0.1) Gecko/2008070206
Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10.5; en-US; rv:1.9.0.1) Gecko/2008070206
Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10.6; en-US; rv:1.9.2.13; ) Gecko/20101203
Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10.6; en-US; rv:1.9.2.3) Gecko/20100402 Prism/1.0b4
Mozilla/5.0 (Macintosh; U; Intel Mac OS X; en-US; rv:1.8.1.11) Gecko/20071127
[...]

PR #1573, opened 2017-09-06
Improving SeaMonkey detection

### LINKED ISSUE #1571: Missing Seamonkey Version (2.48)
2.48 was released on July 31: https://www.seamonkey-project.org/releases/

Unfortunately, since the version was missing, browscap is detecting UAs like this one as Firefox:

>Mozilla/5.0 (Windows NT 10.0; WOW64; rv:51.0) Gecko/20100101 Firefox/51.0 SeaMonkey/2.48

I'm planning to add a generic file for seamonkey so that it should pick new versions up as Seamonkey until the version can be added.

jaydiablo:
Fixes #1571

Adding a SeaMonkey Generic file to catch invalid/unknown versions.

Adding version 2.48 that was released in July.

Split up the SeaMonkey files a little bit to deal with certain platforms not being supported after certain versions, based on information here: https://www.seamonkey-project.org/releases/legacy

Removed some versions that weren’t ever released (can be seen in this table: https://en.wikipedia.org/wiki/SeaMonkey#Release_history)  I did end up leaving 1.5 (due to discussion in #961), but marked it as an Alpha release.

Removed old versions from the standard file (generic file will still catch these as SeaMonkey, but won’t have a version). I made the Standard cut-off at version 2.30 which was released in 2014 (https://www.seamonkey-project.org/releases/2.30).

Also some various platform improvements/additions.

I added a bunch of tests from UAs in our logs. Many of these were already detected as SeaMonkey before these changes, but some of them were not.


tests/fixtures/issues/issue-865.php
```diff
@@ -6149,59 +6149,6 @@
         'lite' => false,
         'standard' => true,
     ],
-    'issue-865-DQ' => [
```
jaydiablo: I removed this test as it is now caught by the SeaMonkey generic file (invalid version number for SeaMonkey currently).

PR #1579, opened 2017-09-08
update tests

### LINKED ISSUE #1544: `Device_Type` in Lite file but not Standard
I ran into this when making some changes for #1528 and wanted to verify that this is the expected behavior.

I have some tests failing due to the `Device_Type` property being a different value than what is expected, but only the Lite tests fail (Full and Standard pass fine).  Both the Lite and Standard files have the same `Device_Type` here, but `Device_Type` isn't available in the Standard file, due to this function:

https://github.com/browscap/browscap/blob/master/src/Browscap/Data/PropertyHolder.php#L162

But is available in the Lite file:

https://github.com/browscap/browscap/blob/master/src/Browscap/Data/PropertyHolder.php#L146

I assume this is expected, since that's how the code is, but I wanted to double check. It seems odd to me that a "lesser" file (the Lite file) contains more properties than the Standard file.

mimmi20:
While trying to remove some properties from the tests, I discovered the in standard mode only the standard mode properties were checked, not the lite mode properties.

To fix this, I wanted to use the filters, who are used to create the output.

I removed the `setLogger` function and added the logger into the constructors. I also removed all fluent interfaces I found.

This PR requires PR #1580 and #1583 to be merged before.
This PR fixes #1544.


.php_cs.cache
```diff
@@ -0,0 +1 @@
...p":257105995,"tests\\BrowscapTest\\Formatter\\JsonFormatterTest.php":-784186953,"tests\\BrowscapTest\\Formatter\\PhpFormatterTest.php":-391938518,"tests\\BrowscapTest\\Formatter\\XmlFormatterTest.php":-2128263355,"tests\\BrowscapTest\\Generator\\BuildCustomFileGeneratorTest.php":461253061,"tests\\BrowscapTest\\Generator\\BuildGeneratorTest.php":-570740156,"tests\\BrowscapTest\\Generator\\Helper\\BuildHelperTest.php":-646473112,"tests\\BrowscapTest\\Helper\\CollectionCreatorTest.php":-1370076622,"tests\\BrowscapTest\\Writer\\CsvWriterTest.php":1101898372,"tests\\BrowscapTest\\Writer\\Factory\\CustomWriterFactoryTest.php":983023773,"tests\\BrowscapTest\\Writer\\Factory\\FullCollectionFactoryTest.php":-1713893646,"tests\\BrowscapTest\\Writer\\Factory\\FullPhpWriterFactoryTest.php":-1270777302,"tests\\BrowscapTest\\Writer\\IniWriterTest.php":-793353956,"tests\\BrowscapTest\\Writer\\JsonWriterTest.php":361816348,"tests\\BrowscapTest\\Writer\\WriterCollectionTest.php":1429067174,"tests\\BrowscapTest\\Writer\\XmlWriterTest.php":-1628250298,"tests\\UserAgentsTest\\FullTest.php":1551869573,"tests\\UserAgentsTest\\LiteTest.php":-459569176,"tests\\UserAgentsTest\\StandardTest.php":727140160}}
```
jaydiablo: This file can cause weird merge conflict issues when merging/rebasing, fyi.

It's nice to have, because it speeds up the php-cs-fixer run quite a bit, but it effectively has to be wiped and re-built if it's ever in a merge conflict.

We haven't figured out a good way to deal with this, and I'm not sure there is other than wiping/rebuilding on conflict, or not committing at all.

How do other open source projects deal with this?

.travis.yml
```diff
@@ -50,6 +50,14 @@ jobs:
       php: 7.1
       env: COMPOSER_FLAGS=""
       script: php -n -d memory_limit=768M vendor/bin/phpcs
+    - stage: check coding style
```
jaydiablo: Do both phpcs and php-cs-fixer need to be used?

phpcs.xml
```diff
@@ -2,13 +2,22 @@
 <ruleset name="PHP_CodeSniffer">
```
jaydiablo: Again, is there a reason to use both phpcs and php-cs-fixer?

Seems that php-cs-fixer does all of these checks/fixes no?

tests/phpunit-compare.xml.dist
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jaydiablo: This file is the same as the "integration" one from #1581 no?

Needs to be removed from this PR?

.travis.yml
mimmi20: Not really. In an other project I had situations where phpcs complaint and cs-fixer did not, so I use both. With the cs-fixer I can check more things, but I did not really search for the rules to do the same things with phpcs.

Which one do you prefer?

.php_cs.cache
mimmi20: I can remove this file and add it to .gitignore.

mimmi20: I have seens some projects which do have this file in the repo, but others have. But I dont know how they handle conflicts.

jaydiablo: Do you think it's important to have the cache file for Travis?

.travis.yml
jaydiablo: If you think that php-cs-fixer can catch the important coding style issues that you're using phpcs to detect, then I'd say move to php-cs-fixer since it's nice that it can correct a lot of issues automatically.

I don't know that it can catch everything that phpcs checks for though, probably depends on the rule set that browscap is using (in my own experience, we use phpcs in the build to report on code style violations (but don't fail the build on them, since a lot of them are hard to fix across the board, like line-length), and we also use it to lint before creating a PR/Diff.  But we also use php-cs-fixer to automatically fix anything it catches (but don't run this in the build)).

phpcs.xml
mimmi20: file removed

tests/phpunit-compare.xml.dist
mimmi20: file removed

.travis.yml
mimmi20: codesniffer removed

.php_cs.cache
mimmi20: file removed and ignored

.gitignore
```diff
@@ -7,6 +7,9 @@ composer.phar
 test*.txt
 /coverage
 coverage.json
+coverage-full.json
```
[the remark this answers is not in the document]
mimmi20: solved

jaydiablo:
@mimmi20 is it possible to have a PR that is just the php-cs-fixer changes, without any other code modifications?

Is it too much work to extract that bit?

There's just a lot to look through here. I think it's mostly just php-cs-fixer changes, but it would be easier to identify the code/logic changes you've made without the php-cs-fixer changes scattered about.

mimmi20:
The coding style changes are separated into PR #1583.

PR #1580, opened 2017-09-08
move tests

mimmi20:
As preparation for PR #1579 I moved all comparasation tests out of the fixtes directory. This helps for checking coding style in PR #1579 and helps preventing merge conflicts in that branch.

This PR has to be merged before PR #1579.


jaydiablo:
I noticed I have write access now, so I could've merged this, just didn't want to overstep any boundaries. :)

PR #1581, opened 2017-09-11
split tests

mimmi20:
To all tests I added the full keyword like the already existing lite and standard keywords. With this it is possible to disable test in the full mode to test differences in the standard or lite mode. I splitted the comparasation test in three pieces.

This PR is required for PR #1579.


.travis.yml
```diff
@@ -25,35 +24,50 @@ before_install:
 install: travis_retry composer update --optimize-autoloader --prefer-dist --prefer-stable --no-progress --no-interaction -vv $COMPOSER_FLAGS
 
 script:
-  - composer validate --strict
```
jaydiablo: You had asked that I add --strict when I added the validate command, any particular reason you're removing this here?

tests/UserAgentsTest/FullTest.php
```diff
@@ -0,0 +1,240 @@
...
+            $coverageProcessor = new Processor(__DIR__ . '/../../resources/user-agents/');
+            $coverageProcessor->process(self::$coveredPatterns);
+            $coverageProcessor->write(__DIR__ . '/../../coverage-full.json');
+        }
+    }
+
+    /**
+     * @return array[]
+     */
+    public function userAgentDataProvider()
```
jaydiablo: Should this have the return type hint ` : array` or is that part of #1579?

tests/UserAgentsTest/LiteTest.php
```diff
@@ -0,0 +1,227 @@
...
+     */
+    private static $filter = null;
+
+    public static function setUpBeforeClass()
+    {
+        // First, generate the INI files
+        $buildNumber    = time();
+        $resourceFolder = __DIR__ . '/../../resources/';
+
+        self::$buildFolder = __DIR__ . '/../../build/browscap-ua-test-' . $buildNumber . '/build/';
```
jaydiablo: Before these changes, a run that would run the Lite, Standard and Full tests would write to the same directory, since the `$buildNumber` was shared in all runs, with this change, they could potentially write to different directories, is this desired?

If so, would it be helpful to mark the directories differently for each run, i.e. ?

`browscap-ua-test-full-' . $buildNumber . '`
`browscap-ua-test-standard-' . $buildNumber . '`
`browscap-ua-test-lite-' . $buildNumber . '`

tests/phpunit-compare.xml.dist
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jaydiablo: I've always thought it was odd that the useragent tests were marked by the "COMPARE" flag. Is this because these tests evolved from the `compare` tests that `browscap-php` runs (where it compares `browscap-php` and `get_browser`)?

Should this flag/phpunit.xml file be named something else for this suite?

mimmi20: The comparasation is made between the results written in out 15000 testcases and the real result returned the browscap-php.
This separate test case is nessesary to move out the test cases from the BrowscapTest directory to make code style checks faster for the other tests.

If you have a better name for this test step, please tell me.

tests/UserAgentsTest/FullTest.php
mimmi20: I will add all the type hint in PR #1579.

.travis.yml
mimmi20: No.

mimmi20: The strict option is available again. Just of Copy mistake.

tests/UserAgentsTest/LiteTest.php
mimmi20: Folders are named now, like you suggested

tests/phpunit-compare.xml.dist
jaydiablo: It's not that big of a deal really.

But a couple of suggestions:

`phpunit-integration.xml.dist`
`phpunit-useragents.xml.dist`

I think I prefer the first one, since they're really acting as integration tests.

jaydiablo:
Oh, I also was looking at the codecov report, and it doesn't have the #lite and #standard tags in it, like on this file: https://codecov.io/gh/browscap/browscap/src/b980912f597b950512dafd24f2014caad4abda5f/resources/user-agents/browsers/chrome/chrome-android-43-on.json

In the Travis log, it doesn't seem like the Lite and Full files uploaded:

https://travis-ci.org/browscap/browscap/jobs/274238188#L760
https://travis-ci.org/browscap/browscap/jobs/274238187#L913

I think that might be that same issue that I mentioned in a previous PR, where the Writer has to be passed to the Property Holder in order for the pattern ids to be added to the file.  I think you're making this change in #1579 (https://github.com/browscap/browscap/pull/1579/files#diff-70eb4236ebbed1f190d77c9fde7f1bc5R142 and https://github.com/browscap/browscap/pull/1579/files#diff-70eb4236ebbed1f190d77c9fde7f1bc5R159 and https://github.com/browscap/browscap/pull/1579/files#diff-e0e176fbe632dd3ebf6d08584192a9ffR92 for Lite).  I feel that the Standard property check needs it too, like here (https://github.com/browscap/browscap/pull/1579/files#diff-70eb4236ebbed1f190d77c9fde7f1bc5R193) but it's not present in that diff, and it seems that the Standard coverage file uploaded ok there (https://codecov.io/gh/browscap/browscap/src/68a608ed0c9a4bbc70da63eb60bf9191539cdda0/resources/user-agents/browsers/chrome/chrome-android-43-on.json  Notice the three tag buttons in the upper right of the file view).  Ah, I think it's because it's checking if the property is a Lite Mode property first here: https://github.com/browscap/browscap/pull/1579/files#diff-2aea9981e7bbeea7643cac2c22676c90R92

Should these changes be brought into this PR so the coverage reports upload properly?

.travis.yml
```diff
@@ -4,7 +4,6 @@ sudo: false
 php:
   - 7.1
   - 7.2
-  - nightly
```
jaydiablo: Is the removal of nightly here intentional?

It still remains in the `allow_failures` section.

mimmi20: It is actual not possible to install the project on PHP 7.2 and Nightly when the CS-Fixer is included. On lines 36-39 the project is installed without php-version-check in composer.

Removing nightly in the matrix saves 2 jobs.

PR #1591, opened 2017-09-22
Cleaning up Openwave Mobile patterns some

### LINKED ISSUE #1590: Openwave Mobile Browser, more cleanup
I ran into a couple of the useragents in the tests that appear to be detected improperly (either missing a version, or were detected as something else).

While looking at these, I decided to dump some more "Openwave Mobile Browser" UAs and make sure any changes I make don't affect those.

I also noticed that many of the openwave files have a lot of patterns that look like they were meant for a particular device, but don't have a device defined.  I'm planning to remove those (unless tests hit them, which in that case I'll add the necessary device).

jaydiablo:
Fixes #1590.

I removed a bunch of device specific patterns that had no devices attached (if I had a UA that did match any of those, I added a pattern with a device). I also added some more devices to match some UAs I had (made tests from what I needed to).

Also simplified the patterns some in the 6.2 file (I had a lot of UAs touching these patterns, so made it easier to change without breaking any detection).

Added “JAVA” as the default platform for all Openwave browser files (some had it some didn’t).

Made sure the comment for all openwave versions included the version.

Tests added.


tests/UserAgentsTest/FullTest.php
```diff
@@ -164,7 +164,7 @@ public function userAgentDataProvider()
                 continue;
             }
 
-            $tests = require_once $file->getPathname();
```
jaydiablo: I changed these to `require` instead of `require_once` because sometimes I like to run all of the UA tests in a single command, like: `./vendor/bin/phpunit -c tests/phpunit-integration.xml.dist --no-coverage --colors --verbose tests/UserAgentsTest/` and when I did this with the `require_once` the Standard and Lite tests actually would throw notices that `$tests` wasn't an array, and would end up skipping those suites since the data provider had no data to return.

PR #1605, opened 2017-10-03
Cleaning up Yandex Browser detection

### LINKED ISSUE #1604: Yandex Browser cleanup + coverage
I was looking for useragents that were missing a platform version (implying that the platform version needs to be added to the pattern) and during this process I found myself working in the Yandex Browser files, which seemed like they could be separated a bit, and have some tests added for coverage.

PR coming.

jaydiablo:
Fixes #1604

I’ve split the yabrowser files into three different groups; iOS, Android and Desktop, similar to what we’ve done with some other browsers. The main reason for this was so that I could split the files at different versions without having to copy the android devices around so much.

I’ve removed a bunch of versions of the browser that don’t seem to exist (other than recent iOS versions, it’s difficult to find a good source for version history of this browser, but I did grab every Yandex Browser UA I could find and based the removals on the versions from those UAs).

I’ve added platforms where it made sense, and removed support for some older platforms based on Yandex’s and Apple’s sites (i.e., Win XP isn’t supported in most recent version of desktop, and iOS 8.0 is required for most recent version of iOS browser).

I’ve also removed some cases of “Linux_32” that weren’t necessary, and also removed some pattern variations that weren’t used (based on a coverage report I made from the previously mentioned UA list).

I ran these changes and the current copy of `master` against ~7,500 Yandex Browser useragents and confirmed that all are parsed **better** than `master` (have correct browser, browser version, platform version or device (relative to `master`)).

This does reduce the pattern count in the full file some (mainly due to removal of non-existent versions):

master line count: 2,494,894
this branch line count: 2,437,539

master pattern count: 227,274
this branch pattern count: 221,607


tests/issues/issue-635.php
```diff
@@ -30179,9 +30179,9 @@
             'Device_Pointing_Method' => 'touchscreen',
             'Device_Code_Name' => 'A680',
             'Device_Brand_Name' => 'Lenovo',
-            'RenderingEngine_Name' => 'Blink',
```
jaydiablo: This one should actually be WebKit because it's earlier than Chrome 27.  Previous to this PR, the `1.20` version was in the same file as the `13.9`+ patterns, which are on Blink.  In my changes above, this (seemingly android only) version was moved to a file that contains earlier versions of the browser, which all appear to be previous to Chrome 27.

PR #1606, opened 2017-10-04
Restore IniParser implementation which was removed in 72914e0f

asgrim:
`IniParser` got removed in #1583 thinking it wasn't used; however, it's used by the website; so we need to leave it here until it can get ported over (as it probably belongs there anyway).

Without the `IniParser`, I can't release the version (so 6025 is a "dead" version; assuming this works, we'll go straight to 6026)

### DIFF
--- a/src/Browscap/Parser/IniParser.php
+++ b/src/Browscap/Parser/IniParser.php
@@ -1,6 +1,14 @@
 <?php
-declare(strict_types=1);
+/**
+ * This file is part of the browscap package.
+ *
+ * Copyright (c) 1998-2017, Browser Capabilities Project
+ *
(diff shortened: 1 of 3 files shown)


mimmi20:
@asgrim You should be able to remove a release on Github.

asgrim:
Yes, in theory, but once a tag is tagged in git, it should never be untagged (despite what some people do...)

asgrim:
Created browscap/browscap-site#31 to get the `IniParser` moved over to `browscap-site` project.

mimmi20:
@asgrim You dont need to port the `IniParser` to the website. You could use the implementation from browscap-php.

asgrim:
Can you comment on that issue please? :)

PR #1643, opened 2017-10-24
Updating CFNetwork versions, re-structuring CFNetwork Safari some

### LINKED ISSUE #1633: iOS 11 CFNetwork/Darwin detection
see ua-parser/uap-core#278 for more details.

jaydiablo:
Fixes #1633.

I added new CFNetwork versions for iOS 10 and 11 where needed.

I also grabbed ~1500 CFNetwork UAs from wherever I could find them and ran them through browscap, making small changes where necessary to get them to parse properly.

This led me to Safari and Mobile Safari which seemed like they were making some assumptions that didn’t hold up (particularly with mobile safari).

I had a few UAs that had a CFNetwork version for iOS 10/11 but were detecting as Safari 4.0 (due to the build number after `Safari/`). This really isn’t possible, since iOS version and Safari version pretty much match up these days, and it’s not possible to stay on an old version of Safari when you upgrade iOS on your device.

I also took a look at desktop Safari CFNetwork patterns and made the OSX versions match the `Mozilla` patterns earlier in the file (so that there’s no OSX 5 in a version of Safari that we already removed that version from for the “regular” pattern, for example).

I’ve added tests where necessary, and did look over the 1,500 useragents that I mentioned earlier to be sure the CFNetwork version was being detected properly in the agents that didn’t end up as tests.


resources/user-agents/browsers/mercury/mercury-generic.json
```diff
@@ -1,17 +1,17 @@
 {
-  "division": "Mercury",
+  "division": "Mercury Generic",
```
jaydiablo: I added "Generic" here to be more consistent with other files, this did require changing some tests.

resources/user-agents/browsers/mobile-safari/mobile-safari-4-0.json
```diff
@@ -78,132 +78,6 @@
...
-          ]
-        },
-        {
-          "match": "MobileSafari/6??.* CFNetwork/711.5*",
-          "platforms": [
-            "iOS_A_8_4"
-          ]
-        },
-        {
-          "match": "MobileSafari/6??.* CFNetwork/758*",
```
jaydiablo: These are some of the patterns I mention in the description.  It shouldn't be possible to have Safari 4.0 on iOS 9.0 for example.  So for mobile safari I made the versions match the iOS version (and put the patterns in the generic file).

I could move the 4.0 patterns back to this file perhaps. I had originally moved them out since they don't have any `#MAJORVER#` tokens, but that should be fine as long as there's only one version in the file.

Preference?

resources/user-agents/browsers/safari/safari-10-0.json
```diff
@@ -1,7 +1,7 @@
 {
   "division": "Safari #MAJORVER#.#MINORVER#",
-  "versions": ["9.1", 9],
```
jaydiablo: I had to break up these files a little bit for the CFNetwork patterns so they could live in a versioned file that only contains one version.

resources/user-agents/browsers/safari/safari-9-0.json
```diff
@@ -0,0 +1,143 @@
...
+          ]
+        },
+        {
+          "match": "Safari/106??.* CFNetwork/811*(x86_64)*",
+          "platforms": [
+            "OSX_10_12"
+          ]
+        },
+        {
+          "match": "Safari/106??.* CFNetwork/*",
```
jaydiablo: As for these Safari build versions, I did some scouring on google to see if I could match up the different versions I was seeing with a Safari browser version number, I believe I was able to pinpoint these as best as possible (there are some new tests in the 1633 file that hit these).

tests/issues/issue-279.php
```diff
@@ -2049,8 +2049,8 @@
             'RenderingEngine_Version' => 'unknown',
             'RenderingEngine_Maker' => 'Apple Inc',
         ],
-        'lite' => true,
-        'standard' => true,
+        'lite' => false,
```
jaydiablo: Because I'm making the Safari version more dependent on the iOS version, if the Standard or Lite file don't contain the iOS version that I use, then they'll be detected as "Safari Generic", which is why I've changed these to `false` in some cases.

tests/issues/issue-618.php
```diff
@@ -652,14 +652,14 @@
     'issue-618-M' => [
         'ua' => 'MobileSafari/600.1.4 CFNetwork/711.1.16 Darwin/14.0.0',
```
jaydiablo: Here's an example where we shouldn't see Safari 4.0 on iOS 8.1.  The previous patterns were looking at the safari build number to decide if it was Safari 4.0, or 6.0...  but in the UAs I was seeing, that build number doesn't vary that much (for mobile safari), and it was detecting a lot of UAs as 4.0:

```
MobileSafari/600.1.4 CFNetwork/711.0.6 Darwin/14.0.0
MobileSafari/600.1.4 CFNetwork/711.1.12 Darwin/14.0.0
MobileSafari/600.1.4 CFNetwork/711.1.16 Darwin/14.0.0
MobileSafari/600.1.4 CFNetwork/711.3.18 Darwin/14.0.0
MobileSafari/601.1 CFNetwork/758.2.8 Darwin/15.0.0
MobileSafari/601.1 CFNetwork/758.4.3 Darwin/15.5.0
MobileSafari/601.1 CFNetwork/758.5.3 Darwin/15.6.0
MobileSafari/602.1 CFNetwork/808.0.2 Darwin/16.0.0
MobileSafari/602.1 CFNetwork/808.1.4 Darwin/16.1.0
MobileSafari/602.1 CFNetwork/808.2.16 Darwin/16.3.0
MobileSafari/602.1 CFNetwork/808.3 Darwin/16.3.0
MobileSafari/602.1 CFNetwork/811.4.18 Darwin/16.5.0
MobileSafari/602.1 CFNetwork/811.5.1 Darwin/16.6.0
MobileSafari/602.1 CFNetwork/811.5.4 Darwin/16.6.0
MobileSafari/602.1 CFNetwork/811.5.4 Darwin/16.7.0
MobileSafari/604.1 CFNetwork/887 Darwin/17.0.0
```

resources/user-agents/browsers/mobile-safari/mobile-safari-4-0.json
jaydiablo: I moved what I could back to the versioned files, from the generic file.

The others would require that I split up the safari version files some.  Maybe something to do in the future.  I may look to clean up these files some, as there are a lot of iOS versions in these files that probably shouldn't be there. I want to run against some UAs though to confirm.

PR #1657, opened 2017-10-31
Issue 1656

mmorel-35:


mmorel-35:
> UA "Java/1.8.0_05" added more than once, now for key "issue-1656-B", before for key "issue-1597-CR"

issue-1656-B owns the right version, should I update issue-1597-CR or delete it to use only issue-1656-B ?

PR #1662, opened 2017-11-02
Changing travis to use composer install

### LINKED ISSUE #1661: Travis is running "composer update" which ignores the lock file
This project has a `composer.lock` file committed to lock the composer dependencies for all packages at the time the lock file was built, however, the travis build runs `composer update` (https://github.com/browscap/browscap/blob/master/.travis.yml#L24), which ignores the lock file and pulls down the latest dependencies that match the version restrictions defined in `composer.json` which may pull down newer versions of packages than what is defined in the `composer.lock` file, and can cause things to break that didn't break in the developer's local copy, like what we're seeing in #1660.

I believe the travis build should be changed to run `composer install` instead of `composer update` so that it uses the dependencies defined in the lock file, giving us better control of which packages are used in the build (i.e., we wouldn't commit a lock file with an updated `php-cs-fixer` version without running `php-cs-fixer` to update any files that the new version may catch).

jaydiablo:
Fixes #1661

Since this project has the composer.lock file in the repo travis should obey it.

Changing travis to use `composer install` instead of `composer update`. This required removing a couple of composer flags, but when `install` is used, these flags don’t have an effect (which is probably why they’re not available).

I also had to lock php-cs-fixer to 2.7.1 as part of this, as I needed to do a `composer update` to make the lock file match the json file (with the new `prefer-stable` option added), otherwise it would have pulled down 2.7.2 which errors out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ php:
 
 os: linux
 
-env:
-  - COMPOSER_FLAGS=""
-
 cache:
   directories:
     - $HOME/.composer/cache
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -21,7 +20,7 @@ before_install:
   - echo 'opcache.enable_cli=1' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
   - travis_retry composer self-update
 
-install: travis_retry composer update --optimize-autoloader --prefer-dist --prefer-stable --no-progress --no-interaction -vv $COMPOSER_FLAGS
+install: travis_retry composer install --optimize-autoloader --prefer-dist --no-progress --no-interaction -vv $COMPOSER_FLAGS
```
mimmi20: I would prefer to leave the update here, as it prevents from failing builds because of incompatilities between the versions defined in the composer.lock and the versions supporting the php versions the tests running on.

mimmi20: I would prefer to insert the modified line directly into coding style stage.

mimmi20: We also could run the Update without dev-dependencies and load phars for the tools we use here to check.

.travis.yml
```diff
@@ -8,7 +8,6 @@ php:
 os: linux
 
 env:
-  - COMPOSER_FLAGS="--prefer-lowest"
```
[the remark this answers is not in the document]
jaydiablo: ok, it seems that these env vars aren't used for the checkstyle stage anyhow:

https://travis-ci.org/browscap/browscap/builds/295482248

.travis.yml
jaydiablo: I can certainly try that, I just wonder if a composer.lock file should be committed at all for this project if we're concerned about testing against different versions of packages (as a library would typically do).

With the lock file you're basically saying that this application works against this set of dependencies, and has been tested that way, and should use those when installed, but if we're testing against multiple PHP versions which could pull different dependencies (and if one developer does a composer update under one PHP version, but another does one with another PHP version, the lock file could potentially be different between the two versions) then I think perhaps the lock file isn't serving its purpose here, and shouldn't be committed.

IMO, this project **should** be considered an application, which should only run on a particular version of PHP and have a lock file that defines which dependencies are known to work, and should be installed with it.  However, I know that @asgrim does use some of this project in the browscap site as a library. So perhaps it should be considered a library, and we ignore the lock file?

If that's the route we take, then we'll just have to react to these sorts of build failures that are caused by a dependency updating from one day to another, and either fix the code to work with that dependency (but still work with the "lowest"), or change `composer.json`  to lock to a specific version, or range of versions that don't have an issue.

Thoughts?

I can modify this PR to just remove and ignore the lock file, and lock the php-cs-fixer version at 2.7.1 so the build doesn't break until the issue with 2.7.2 can be addressed.

asgrim:
Should be treated as a project imo, so only need to test against versions in `composer.lock`. The site is a bit unique in that we consume it, but we manage that "internally" anyway. So yeah, moving to `composer install` makes sense

mimmi20:
Then the `COMPOSER_FLAGS` env var should be removed completely.

asgrim:
Yup :+1:

PR #1669, opened 2017-11-04
Issue 1635

mmorel-35:
Fix


resources/user-agents/apps/kmail2-generic.json
```diff
@@ -0,0 +1,40 @@
+{
```
jaydiablo: I think it would be nice to move these to their own folder: `resources/user-agents/apps/kmail2/`

PR #1670, opened 2017-11-04
Issue 1618

mmorel-35:
You might want to change the "sortIndex" attribute or other attributes.


tests/issues/issue-1618.php
```diff
@@ -0,0 +1,220 @@
+<?php
+
+return [
+    'issue-1618-A' => [
+        'ua' => 'Evolution/3.28.0',
```
jaydiablo: Shouldn't this parse as Evolution 3.28?

tests/issues/issue-1618.php
```diff
@@ -0,0 +1,220 @@
...
+            'RenderingEngine_Name' => 'unknown',
+            'RenderingEngine_Version' => 'unknown',
+            'RenderingEngine_Maker' => 'unknown',
+        ],
+        'lite' => false,
+        'standard' => false,
+        'full' => true,
+    ],
+    'issue-1618-B' => [
+        'ua' => 'CamelHttpStream/1.0 Evolution/3.28.0',
```
jaydiablo: This one too?

resources/user-agents/apps/evolution-generic.json
```diff
@@ -0,0 +1,27 @@
+{
```
jaydiablo: It may be nice to move these into their own folder: `resources/user-agents/apps/evolution/`

tests/issues/issue-1618.php
mmorel-35: The 3.28 doesn't exists yet

jaydiablo: I think you should leave these out of the test suite then.

I don't generally add tests that are picked up by the generic file, unless they're an "invalid version", in which case we're trying to keep those in the `issue-000-invalid-versions.php` file. A "hasn't been released yet" version doesn't really qualify for that though, because once that version is released, the test will have to be modified, which just adds more maintenance.

PR #1676, opened 2017-11-06
Standardisation of the user-agents arborescence

### LINKED ISSUE #1675: Folder by Browser_Type
Under  "browscap/resources/user-agents/" are four folder representing different browser types.
But it seems that it's a bit messy for now.
How do you feel about having a strict differentiation?
For example 
"resources/user-agents/apps/barca-mail-client.json" and "resources/user-agents/browsers/kontact/kontact-generic.json" would move to 
"resources/user-agents/email-clients/barca-mail-client.json" and "resources/user-agents/email-clients/kontact/kontact-generic.json".

I'm taking the example of "Email Client" but it can be applied with caution to the rest of the browser types.

mmorel-35:
Fix #1675


mmorel-35:
I alsso noticed that Realplayer and stagefright are in media players folder but are qualified as Applications, should'nt they be Multimedia Player ?

mimmi20:
You are right. Could you fix this?

PR #1704, opened 2017-11-18
Adding a bunch of Samsung SM-A500* devices (Galaxy A5 variants)

### LINKED ISSUE #1688: Galaxy A5 not in the DB?
Hey guys, this UA isn't matched, even by full version of DB. IT looks like it must be Samsung Galaxy A5 (2015)
`Mozilla/5 (Linux; Android 6; SM-A500W Build/MMB29M) AppleWebKit/537 (KHTML, like Gecko) Chrome/59 Mobile Safari/537`

jaydiablo:
Fixes #1688

Found as many legit UAs as I could with the variations of the `SM-A500` device (Galaxy A5) and added devices where necessary to properly detect them all.

I didn’t end up adding every device variation I found to every browser/app as I didn’t want to bloat the files, so only added what I had UAs for.

I did try to keep them on the same 3 versions of Android across all files that I added to (unless that file doesn’t support a version).

I compared the results of these changes to `master` on a couple of large sets of UAs and confirmed that the only difference in parsing was the device info.


resources/user-agents/apps/naver/naver-android.json
```diff
@@ -0,0 +1,57 @@
+{
+  "division": "Naver #MAJORVER#.#MINORVER# for Android",
```
jaydiablo: I did end up splitting the Naver android patterns into their own file so I could add some older versions without having to add the android patterns to the naver file that has more versions for iOS.

resources/user-agents/browsers/android-browser/android-browser-4-0.json
```diff
@@ -1276,8 +1287,7 @@
             "A5000": "Sony A5000"
           },
           "platforms": [
-            "Android_4_0",
-            "Android"
```
jaydiablo: I removed "Android" here because it was causing the Lite and Standard files to catch some UAs with `SM-A5000` as this device, which was incorrect.  The UA in the test suite for this pattern is an Android 4.0 UA, so I 'm not sure why the generic Android platform was added here.

resources/user-agents/browsers/maxthon/maxthon-4-1-to-4-3.json
```diff
@@ -8,7 +8,7 @@
     {
       "userAgent": "Maxthon #MAJORVER#.#MINORVER# for Android",
       "platform": "Android",
-      "engine": "WebKit",
```
jaydiablo: I swapped these as my script that I used in #1692 showed that by adding the devices I did to this file caused there to be more Blink patterns than WebKit now.

tests/issues/issue-1413.php
```diff
@@ -43538,8 +43538,8 @@
             'RenderingEngine_Version' => 'unknown',
             'RenderingEngine_Maker' => 'Apple Inc',
         ],
-        'lite' => true,
```
jaydiablo: This is the test that has that Sony A5000 device on Android 4.0 in it, by removing the "Android" platform, it doesn't get picked up by the Lite and Standard files any longer.

PR #1714, opened 2017-11-19
Adding default devices where missing, removing Linux_32 platform

jaydiablo:
Similar to the engine optimization I did earlier, I was curious if the same could be done with devices/platforms.

Platform isn’t as easy since it kind of has to be done with a device combination (don’t want the default platform be Windows and the default device be Macintosh, for example), so I may need to spend more time with that one.

However, I did notice that a lot of files had `Linux_32` platform with the `Linux` platform, so I’ve removed all cases of that.

I also added a default device to useragents where possible in order to take advantage of the line reduction due to parent/child inheritance.

This reduces the number of lines in the `full_php_browscap.ini` file from 2,274,300 to 2,242,623, a savings of 31,677 lines.

No tests had to be modified.


resources/user-agents/browsers/maxthon/maxthon-4-1-to-4-3.json
```diff
@@ -380,20 +369,44 @@
             "Win8_1_x64", "Win8_1_64", "Win8_1_32",
             "Win10_x64", "Win10_64", "Win10_32", "Win10_x64_B", "Win10_64_B", "Win10_32_B", "Windows"
           ]
+        }
+      ]
+    },
+    {
+      "userAgent": "Maxthon #MAJORVER#.#MINORVER# for Mac",
```
jaydiablo: This was something I wanted to test.  By having the default platform/device be similar to what the children are, a lot of lines can be saved when there are a lot of patterns in the UA group. For browsers like this one, it may make sense to split up the desktop patterns into their own UA group (Linux, Mac, Windows).

The biggest downside to this that I see is that we have to duplicate the browser properties in each UA group, but with the browser extraction that @mimmi20 was thinking about doing, there would be a lot less of that, which may make this a more attractive approach to keep the final INI file size smaller.

Thoughts?

For this PR, I only did this for this file and the next two maxthon files.

resources/user-agents/browsers/chrome/headless-chrome.json
```diff
@@ -21,7 +21,7 @@
           "match": "Mozilla/5.0 (#PLATFORM#) applewebkit* (*khtml*like*gecko*) HeadlessChrome Safari/*",
           "device": "Linux Desktop",
           "platforms": [
-            "Linux_64", "Linux_32_x86", "Linux_32", "Linux"
```
mimmi20: Is it possible to remove these two platforms completely (also from the platforms file)?

jaydiablo: I'm not sure yet, I only looked for cases where they were used on the same pattern as the "Linux" platform.

I can take a look though.

jaydiablo: Looks like they can be removed (and another platform that just extended "Linux" but didn't change any properties that wasn't actually used anywhere).  There were a couple of standard specific tests that were testing for "unknown" platform when I changed `Linux_32` to `Linux` in a couple of cases, but I've removed those tests, and changed the `standard` prop on the tests they were derived from to `true`.

Update coming.

PR #1719, opened 2017-11-20
Adding Google Pixel 2 (XL), small inheritance optimization

### LINKED ISSUE #1718: Add Google Pixel 2
Mozilla/5.0 (Linux; Android 8.0.0; Pixel 2 Build/OPD1.170816.025) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.84 Mobile Safari/537.36

Mozilla/5.0 (Linux; Android 8.0; Pixel 2 XL Build/OPD3.170816.012) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.98 Mobile Safari/537.36

jaydiablo:
Fixes #1718.

I searched our logs for these two devices and added the devices to any browsers/apps I found (on Android 8.0 and 8.1, as it seems they’re in the developer preview program for 8.1 and will most likely be upgraded to 8.1 officially once it releases).

Also, while doing this, I noticed that Puffin and Instagram had some iOS and Android UAs in the same Useragent group, so I broke these out into two different groups for both so that they can inherit properties more without having to overwrite on each child. Believe it or not, this actually resulted in a net reduction of 13,727 lines in `full_php_browscap.ini`, even after adding the new platforms/devices for these Pixel 2 devices.

Added tests where necessary for coverage, made sure all UAs I found parsed properly.


tests/issues/issue-1430.php
```diff
@@ -2789,9 +2789,9 @@
             'Device_Pointing_Method' => 'touchscreen',
             'Device_Code_Name' => 'iPhone',
             'Device_Brand_Name' => 'Apple',
-            'RenderingEngine_Name' => 'Blink',
```
jaydiablo: I did end up catching a pattern that was set as `Blink` even though it was an iOS pattern, that's why these tests were changed.

jaydiablo: It wasn't explicitly set as `Blink`, but was in a UA group that was set to `Blink` and didn't have a `WebKit` override. It has been moved to an iOS group now that is set to `WebKit` as the group's engine.

mimmi20 (approved):

PR #1728, opened 2017-11-30
Added Galaxy S7 on Android 7,6

negram:


resources/user-agents/browsers/chrome/chrome-android-43-on.json
```diff
@@ -82,6 +82,7 @@
             "ONEPLUS A3010": "OnePlus A3010",
             "SM-G950F": "Samsung SM-G950F",
             "SM-T820": "Samsung SM-T820",
+            "SM-G930F": "Samsung SM-G930F",
```
jaydiablo: You could create a new group that has just `Android_7_0` and `Android_6_0`, and add this new device there so that it's not defined twice.

It doesn't really change anything in the resulting files when built, but it does help us keep things organized a bit more (It's easy to see at quick glance what android versions are defined for this device).

negram: Done in the next commit (see below)

anna64000:
o yes

PR #1743, opened 2017-12-05
Pixel, Nexus now have Andriod 8.1 onboard

negram:


resources/user-agents/browsers/chrome/chrome-android-43-on.json
```diff
@@ -38,6 +38,7 @@
             "Pixel": "Google Pixel"
           },
           "platforms": [
+            "Android_8_1",
```
jaydiablo: I generally group like major versions of Android, i.e.:

```
"Android_8_1", "Android_8_0",
"Android_7_0"
```

Just FYI though.

PR #1746, opened 2017-12-10
update integration tests, update dependencies

mimmi20:
With this PR intends the following changes:

- move validating the composer.json to its own stage
- add tests against the next version of browscap-php
- update the dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,11 +53,6 @@ jobs:
 
     - stage: composer validate
       php: 7.1
-      before_install:
-        - stty cols 130
-        - phpenv config-rm xdebug.ini || echo "xdebug not available"
-        - echo 'opcache.enable=1' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
-        - echo 'opcache.enable_cli=1' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
       install: skip
       script: composer validate --strict
 
@@ -82,7 +77,6 @@ jobs:
         - travis_retry composer update --optimize-autoloader --prefer-dist --prefer-stable --no-progress --no-interaction -vv $COMPOSER_FLAGS
         - travis_retry composer require --update-with-dependencies --prefer-dist --no-suggest browscap/browscap-php:dev-master
       script: php -n vendor/bin/phpunit -c tests/phpunit-integration.xml.dist --no-coverage --colors --columns 121 --verbose tests/UserAgentsTest/Full4Test.php
-      after_success: bash <(curl -s https://codecov.io/bash) -f coverage-full4.json -F full4
 
     - stage: integration tests
       php: 7.1
@@ -93,7 +87,6 @@ jobs:
         - travis_retry composer update --optimize-autoloader --prefer-dist --prefer-stable --no-progress --no-interaction -vv $COMPOSER_FLAGS
         - travis_retry composer require --update-with-dependencies --prefer-dist --no-suggest browscap/browscap-php:dev-master
       script: php -n vendor/bin/phpunit -c tests/phpunit-integration.xml.dist --no-coverage --colors --columns 121 --verbose tests/UserAgentsTest/Standard4Test.php
-      after_success: bash <(curl -s https://codecov.io/bash) -f coverage-standard4.json -F standard4
 
     - stage: integration tests
       php: 7.1
@@ -104,7 +97,6 @@ jobs:
         - travis_retry composer update --optimize-autoloader --prefer-dist --prefer-stable --no-progress --no-interaction -vv $COMPOSER_FLAGS
         - travis_retry composer require --update-with-dependencies --prefer-dist --no-suggest browscap/browscap-php:dev-master
       script: php -n vendor/bin/phpunit -c tests/phpunit-integration.xml.dist --no-coverage --colors --columns 121 --verbose tests/UserAgentsTest/Lite4Test.php
-      after_success: bash <(curl -s https://codecov.io/bash) -f coverage-lite4.json -F lite4
 
     - stage: integration tests
       php: 7.1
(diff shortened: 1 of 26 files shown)


.travis.yml
```diff
@@ -11,17 +11,24 @@ cache:
   directories:
     - $HOME/.composer/cache
 
+stages:
+  - composer validate
+  - test
+  - test with coverage
+  - check coding style
```
jaydiablo: Would it be nice to move some of the "faster" steps to run before the slower ones (run test, test with coverage and integration tests last)?

If the build is going to fail, would be nice if it fails fast.

mimmi20: changed

.travis.yml
```diff
@@ -33,34 +40,74 @@ jobs:
...
       php: 7.1
       script: php -n -d memory_limit=768M vendor/bin/php-cs-fixer fix --dry-run -vv
     - stage: static code analysis
       php: 7.1
       script: vendor/bin/phpstan analyse -l 7 -c phpstan.neon --autoload-file=vendor/autoload.php --memory-limit=768M --no-progress src tests/UserAgentsTest tests/BrowscapTest tests/fixtures
-    - stage: integration test
+    - stage: integration tests
+      php: 7.1
+      env:
+        - parser="browscap-php 4.x@dev"
```
jaydiablo: Should these be added to an allow_failure block in travis?

Or are we expecting that a fail in the 4.x parser should break a build?

mimmi20: I would like it, that we handle version 4 the same as the current version.

.travis.yml
```diff
@@ -31,44 +38,94 @@ jobs:
...
-      script: vendor/bin/phpunit -c phpunit.xml.dist --colors --columns 121 --verbose --coverage-text --coverage-clover=coverage.clover
+      script: vendor/bin/phpunit -c phpunit.xml.dist --colors --columns 117 --verbose --coverage-text --coverage-clover=coverage.clover
       after_success:
         - wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover
-        - bash <(curl -s https://codecov.io/bash) -f coverage.clover
+        - bash <(curl -s https://codecov.io/bash) -f coverage.clover -F phpunit
+
+    - stage: composer validate
+      php: 7.1
+      before_install:
```
jaydiablo: Is all this necessary for the `composer validate` stage?

.travis.yml
```diff
@@ -31,44 +38,94 @@ jobs:
...
+    - stage: integration tests
+      php: 7.1
+      env:
+        - parser="browscap-php 4.x@dev"
+        - TEST_SET="full"
+      install:
+        - travis_retry composer update --optimize-autoloader --prefer-dist --prefer-stable --no-progress --no-interaction -vv $COMPOSER_FLAGS
+        - travis_retry composer require --update-with-dependencies --prefer-dist --no-suggest browscap/browscap-php:dev-master
+      script: php -n vendor/bin/phpunit -c tests/phpunit-integration.xml.dist --no-coverage --colors --columns 121 --verbose tests/UserAgentsTest/Full4Test.php
+      after_success: bash <(curl -s https://codecov.io/bash) -f coverage-full4.json -F full4
```
jaydiablo: I don't think it's important to run coverage for 3.x and 4.x runs, we should just pick one.

phpstan.neon
```diff
@@ -1,6 +1,10 @@
 parameters:
   excludes_analyse:
     - */tests/*/data/*
+    - */Full2Test.php
```
jaydiablo: Is there a `Full2Test.php`?

phpstan.neon
```diff
@@ -1,6 +1,10 @@
...
+    - */Full4Test.php
```
jaydiablo: Why exclude these?

mimmi20: In v4 the interface changed (removed setters, parameter in contructor). Phpstan will complain this.

.travis.yml
mimmi20: I cleaned this up a bit.

.travis.yml
mimmi20: changed

phpstan.neon
mimmi20: No, I remoed this line.

jaydiablo:
One other thing I noticed with these changes, a lot of times, when I'm running the integration tests manually, and I want to run Full, Standard & Lite tests all at once, I use this command:

```
./vendor/bin/phpunit -c tests/phpunit-integration.xml.dist ./tests/UserAgentsTest/
```

But with these changes, that won't work quite right because the `*4Test.php` files won't load properly (since composer can only load one version at a time).

Should we consider moving the 3x and 4x integration test classes to their own folders?  Then I could run something like this:

```
./vendor/bin/phpunit -c tests/phpunit-integration.xml.dist ./tests/UserAgentsTest/browscap-php-3/
```

To run just the 3.x tests, or:

```
./vendor/bin/phpunit -c tests/phpunit-integration.xml.dist ./tests/UserAgentsTest/browscap-php-4/
```

to run the 4.x tests.

jaydiablo:
Codecov didn't update with the latest commit, perhaps something to do with CircleCI being in there?

jaydiablo:
So why the addition of CircleCI?

Planning to replace Travis?

mimmi20:
No. I wanted to try CircleCi to run the tests with the native get_browser function (in another branch) because CircleCi is faster than Tavis.

.travis.yml
```diff
@@ -31,44 +38,85 @@ jobs:
...
         - stty cols 130
         - echo 'opcache.enable=1' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
         - echo 'opcache.enable_cli=1' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
         - travis_retry composer self-update
-      script: vendor/bin/phpunit -c phpunit.xml.dist --colors --columns 121 --verbose --coverage-text --coverage-clover=coverage.clover
+      script: vendor/bin/phpunit -c phpunit.xml.dist --colors --columns 117 --verbose --coverage-text --coverage-clover=coverage.clover
       after_success:
-        - wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover
-        - bash <(curl -s https://codecov.io/bash) -f coverage.clover
+        - bash <(curl -s https://codecov.io/bash) -f coverage.clover -F phpunit
```
jaydiablo: Ah, I think this has come up before. I don't think it's worth flagging the PHPUnit tests with the "phpunit" flag, since they really won't ever have cross-coverage with the "full"/"lite"/"standard" flags, and just pollutes the codecov UI a little bit.

mimmi20: fixed

mimmi20:
Thanks

PR #1726, opened 2017-11-29
add Bing Search App and devices

mimmi20:
Fixes #1387.

This PR requires PR #1746 to be merged before.
This PR also requires PR #1765 to be merged before because there are similar changes.

PR #1754, opened 2017-12-12
add check-duplicate-tests command

mimmi20:
This PR is required for PR #1726.

### DIFF
--- a/src/Browscap/Helper/IteratorHelper.php
+++ b/src/Browscap/Helper/IteratorHelper.php
@@ -1,0 +1,66 @@
+<?php
+declare(strict_types = 1);
+namespace Browscap\Helper;
+
+use Psr\Log\LoggerInterface;
+
+class IteratorHelper
+{
+    /**
+     * @param LoggerInterface $logger
+     * @param string          $testKey
+     *
+     * @return array
+     */
+    public function getTestFiles(LoggerInterface $logger, string $testKey = 'full') : array
+    {
+        $data = [];
+
+        $checks          = [];
+        $sourceDirectory = __DIR__ . '/../../../tests/issues/';
+        $iterator        = new \RecursiveDirectoryIterator($sourceDirectory);
+
+        $errors = [];
+
+        foreach (new \RecursiveIteratorIterator($iterator) as $file) {
+            /** @var $file \SplFileInfo */
+            if (!$file->isFile() || 'php' !== $file->getExtension()) {
+                continue;
+            }
+
+            $tests = require $file->getPathname();
+
+            foreach ($tests as $key => $test) {
+                if (isset($data[$key])) {
+                    $logger->error('Test data is duplicated for key "' . $key . '"');
+                    $errors[] = 'Test data is duplicated for key "' . $key . '"';
+                }
+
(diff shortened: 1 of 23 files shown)


src/Browscap/Helper/IteratorHelper.php
```diff
@@ -0,0 +1,66 @@
...
+            if (!$file->isFile() || 'php' !== $file->getExtension()) {
+                continue;
+            }
+
+            $tests = require $file->getPathname();
+
+            foreach ($tests as $key => $test) {
+                if (isset($data[$key])) {
+                    $logger->error('Test data is duplicated for key "' . $key . '"');
+                    $errors[] = 'Test data is duplicated for key "' . $key . '"';
```
jaydiablo: To reduce some of this duplication in the messages, I think it might be nice to set an `$error` variable or something like that and use that variable in both the `$logger->error` call and in the `$errors[]` push.

```php
$error = 'Test data is duplicated for key "' . $key . '"';
$logger->error($error);
$errors[] = $error;
```

Same for the two below as well.

src/Browscap/Helper/IteratorHelper.php
```diff
@@ -0,0 +1,66 @@
...
+                }
+
+                if (!array_key_exists($testKey, $test)) {
+                    $logger->error(
+                        '"full" keyword is missing for  key "' . $key . '"'
```
jaydiablo: 'full' here should be `$testKey`.

src/Browscap/Helper/IteratorHelper.php
jaydiablo: Also, I just realized that since the previous code was throwing an exception here, this doesn't behave the same way as it used to (I don't think that's a bad thing, just wanted to bring it up). Now if there are 5 cases of this particular error in the test suite, this will collect all 5 cases of it and store in the `$errors` array, and the methods in the test classes will output all 5 cases.  Previously, it would stop on the first encountered case.  This new approach may be more desirable though, what do you think?

src/Browscap/Helper/IteratorHelper.php
mimmi20: fixed

src/Browscap/Helper/IteratorHelper.php
mimmi20: I removed the duplicate definition of error messages as you suggested.

mimmi20: Its better now, I think. But I dont remember why I didnt just catch the exception in the new command.

jaydiablo:
This looks fine to me, but I know that @asgrim usually has something to say about the docblocks (and some of them changed here, I assume due to a php-cs-fixer run, or even correcting some phpstan complaints?).

Have time to look over this one @asgrim?

asgrim:
What's going on with coverage being red? Is that just because code coverage went down..?

mimmi20:
The new command and rhe new helper are not tested. Because of that the coverage went down.

jaydiablo:
Yeah, total coverage dropped a bit (0.03%) and coverage of touched files was less than project's total coverage (49.81% target, 12.19% of touched files).

Issue #1433, opened 2017-05-30
Remove deprecated properties from tests?


jaydiablo:
I noticed that the list of deprecated properties defined in the Wiki (https://github.com/browscap/browscap/wiki/Resource%3A-User-Agents-Database#deprecated-properties) are still tested for.

As the test suite grows, it may be nice to not assert on these properties anymore, which could see some performance benefits. 

Here's a full run against a branch I'm working on without any modifications to which properties are asserted.

```
Time: 2.77 minutes, Memory: 467.63MB

OK (26652 tests, 1336853 assertions)
```

Here's the same test suite, but with a small bit of test code added to the test runner to ignore that list of deprecated properties (in each file version's run):

```php
$oldProps = ['AolVersion' => 1, 'Crawler' => 1, 'isMobileDevice' => 1, 'isTablet' => 1, 'Win64' => 1, 'Win32' => 1, 'Win16' => 1];

if (isset($oldProps[$propName])) {
    continue;
}
```

```
Time: 2.57 minutes, Memory: 467.63MB

OK (26652 tests, 1125377 assertions)
```

Not an astronomical reduction in execution time, I'll admit (it may reduce a tad more if the properties are removed from the fixtures entirely), perhaps more importantly though, it would reduce some of the maintenance when building/modifying the test fixtures.

Thoughts?

mimmi20:
These properties are defined in the sources too. All of these properties (expect AolVersion) can be generated from other properties (e.g. Crawler from Browser_Type, isMobileDevice/isTablet from Device_Type).
As long the properties are defined in the sources and not generated, we should test them.

jaydiablo:
Is there a plan/timeline to start removing them from the sources?

Should the ones that can be generated be removed first (and code be added that generates them)?

mimmi20:
There is no timeline yet.

mimmi20:
@jaydiablo I would like to extract browser related properties (e.g. `Browser`, `Browser_Maker`, `Browser_Type`, `Frames`, ...) from the useragents into a separate file in the first step. Do you have new Browsers to add? I would wait until these are merged.

mimmi20:
Getting this change working to build browscap files would need some days.

jaydiablo:
I have been running into a fair amount of new browsers as I dig through some useragents for #1448.

I can take a break though if you need.

I don't fully understand what you mean by extracting the browser related properties into a separate file.

[...]

PR #1765, opened 2017-12-23
extract browser related properties

mimmi20:
With this PR browser related properties are moved to their own file and removed them from the useragents section.

As a side effect caused by the script I used to change all source files, the properties inside the sections now are sorted, the versions and the platforms are sorted too.

@asgrim, @jaydiablo Would you like to have this script as a command (in a seperate PR)? 

This PR requires PR #1746 to be merged before.


mmorel-35:
The browser file is really huge. Would it be possible to split it a way that make maintenance easy?

tests/issues/issue-996.php
```diff
@@ -111,7 +109,6 @@
             'JavaScript' => true,
             'VBScript' => false,
             'JavaApplets' => false,
-            'isSyndicationReader' => false,
```
jaydiablo: Why is this property being removed from tests? Is it one of the deprecated properties we're removing as part of #1433?

Are the tests modified in any other way? There are a lot of files to expand here to see what was modified, especially if the only change is the removal of this property.

mimmi20: >  Is it one of the deprecated properties we're removing as part of #1433?

Yes it is.

> Are the tests modified in any other way?

There are some test cases where I added a seperate tests for the lite mode. (These are the test files with added lines)

mimmi20: >> Is it one of the deprecated properties we're removing as part of #1433?

Removing this property was a mistake. I looked not properly enough at the deprecated list.

tests/issues/issue-000-invalid-versions.php
```diff
@@ -316,146 +802,490 @@
...
             'RenderingEngine_Name' => 'Gecko',
             'RenderingEngine_Version' => 'unknown',
+            'RenderingEngine_Maker' => 'Mozilla Foundation',
         ],
         'lite' => false,
         'standard' => false,
         'full' => true,
     ],
-    // Same as above
+    // Firefox 4.0 and up don't suppport OSX 10.4 (https://www.mozilla.org/en-US/firefox/4.0/system-requirements/)
```
jaydiablo: This comment got moved.

tests/issues/issue-000-invalid-versions.php
```diff
@@ -8,12 +8,39 @@
         'properties' => [
             'Comment' => 'Firefox Generic',
             'Browser' => 'Firefox',
+            'Browser_Type' => 'Browser',
```
jaydiablo: I don't think we should add all of these properties to the tests in this file.

tests/issues/issue-000-invalids.php
```diff
@@ -1,60 +1,395 @@
 <?php
```
jaydiablo: Please revert the changes to this file.

tests/issues/issue-000-invalid-versions.php
mimmi20: reverted

tests/issues/issue-000-invalid-versions.php
mimmi20: reverted

tests/issues/issue-000-invalids.php
mimmi20: reverted

src/Browscap/Data/Expander.php
```diff
@@ -71,6 +72,7 @@ private function resetPatternId() : void
             'useragent' => '',
             'platform' => '',
             'device' => '',
+            'browser' => '',
```
jaydiablo: I'm not sure this is necessary to have the browser (or the engine) in the pattern id.

They're not used in the `getPatternId` method, and not included in coverage reports.

If you really do think we should add them (because you want to expand coverage to process the browser and engine files), then you should make sure you add "engine" here too (since you're setting it below).

src/Browscap/Data/Expander.php
```diff
@@ -160,6 +168,23 @@ private function parseUserAgent(UserAgent $uaData, bool $lite, bool $standard, i
             $deviceProperties = [];
         }
 
+        if (null !== $uaData->getBrowser()) {
+            $this->patternId['browser'] = $uaData->getBrowser();
```
jaydiablo: Again, I don't think this is necessary.

src/Browscap/Data/Expander.php
```diff
@@ -237,8 +270,9 @@ private function parseChildren(string $ua, array $uaDataChild, bool $lite = true
                 $uaBase = str_replace('#PLATFORM#', $platformProperties->getMatch(), $uaDataChild['match']);
 
                 if (array_key_exists('engine', $uaDataChild)) {
-                    $engine           = $this->collection->getEngine($uaDataChild['engine']);
-                    $engineProperties = $engine->getProperties();
+                    $engine                    = $this->collection->getEngine($uaDataChild['engine']);
+                    $engineProperties          = $engine->getProperties();
+                    $this->patternId['engine'] = $uaDataChild['engine'];
```
jaydiablo: Not necessary (imo), but if you add here, need to make sure it's in the reset array.

src/Browscap/Data/Expander.php
```diff
@@ -277,8 +327,9 @@ private function parseChildren(string $ua, array $uaDataChild, bool $lite = true
             $properties = ['Parent' => $ua, 'lite' => $lite, 'standard' => $standard];
 
             if (array_key_exists('engine', $uaDataChild)) {
-                $engine           = $this->collection->getEngine($uaDataChild['engine']);
-                $engineProperties = $engine->getProperties();
+                $engine                    = $this->collection->getEngine($uaDataChild['engine']);
+                $engineProperties          = $engine->getProperties();
+                $this->patternId['engine'] = $uaDataChild['engine'];
```
jaydiablo: Another case.

src/Browscap/Data/Factory/BrowserFactory.php
```diff
@@ -0,0 +1,35 @@
...
+declare(strict_types = 1);
+namespace Browscap\Data\Factory;
+
+use Assert\Assertion;
+use Browscap\Data\Browser;
+
+class BrowserFactory
+{
+    /**
+     * Load a engines.json file and parse it into the platforms data array
```
jaydiablo: Not loading engines.json

src/Browscap/Data/Factory/UserAgentFactory.php
```diff
@@ -10,16 +10,11 @@ class UserAgentFactory
...
      * @param bool    $isCore
      *
-     * @throws \RuntimeException If the file does not exist or has invalid JSON
-     * @throws \LogicException   If required attibutes are missing in the division
-     * @throws \LogicException
-     *
      * @return UserAgent[]
      */
-    public function build(array $userAgentsData, array $versions, bool $isCore) : array
+    public function build(array $userAgentsData, bool $isCore) : array
```
jaydiablo: Since the signature is changing here, did you search for all cases where this would have been called?

Also, should be documented in the Readme?

src/Browscap/Data/Validator/DivisionDataValidator.php
```diff
@@ -413,6 +427,26 @@ public function checkDeviceProperties(array $properties, string $message) : void
         }
     }
 
+    /**
+     * checks if device properties are set inside a properties array
```
jaydiablo: "checks if **browser** properties"

tests/BrowscapTest/Data/DataCollectionTest.php
```diff
@@ -26,21 +26,6 @@ public function setUp() : void
         $this->object = new DataCollection($logger);
     }
 
-    private function getPlatformsJsonFixture()
```
jaydiablo: I assume these are unused now (I think I recall you removing these fixture files a while back).

src/Browscap/Data/Expander.php
mimmi20: fixed

src/Browscap/Data/Expander.php
mimmi20: fixed

src/Browscap/Data/Expander.php
mimmi20: fixed

src/Browscap/Data/Expander.php
mimmi20: fixed

src/Browscap/Data/Factory/BrowserFactory.php
mimmi20: fixed

src/Browscap/Data/Validator/DivisionDataValidator.php
mimmi20: fixed

src/Browscap/Data/Factory/UserAgentFactory.php
mimmi20: added to Readme

tests/BrowscapTest/Data/DataCollectionTest.php
mimmi20: yes, they were unused.

jaydiablo:
@mimmi20 regarding that script, I think it could be useful for sure.

I have a number of different scripts I use when doing updates, I could look at moving them into commands as well if they seem useful to others.

PR #1774, opened 2018-01-23
Adding missing browser versions

jaydiablo:
This adds missing versions for UAs from 2017/12 and 2018/01 (so far).

For the most part I’m just adding new versions to the versions array, but in a couple of cases I added some new versions of iOS as well to be sure the agent parsed properly.

As for UC Browser changes, I started splitting the files a little bit into ios/desktop/other (android in other for now) because the new desktop version (7.0) needed to be added, but I didn’t want all of the 7.x versions in the current 7.x file to be included, since there are no such versions yet.

There is some weirdness with UC Browser 10/11 though. These versions are defined for Windows Desktop platforms, but they’re higher than the current version of the UC Browser for Windows available at http://www.ucweb.com/desktop/.  I don’t know if this is because there was an older desktop version that used a slightly different UA, or if these are fake/spoofed, or something else (desktop mode in mobile browser?).

More research is necessary there I believe, but this PR doesn’t really change any of those patterns, just moves some of them to new files, and leaves the rest where they were.

Tests added where necessary/possible.  Did confirm that the list of UAs I had all now have a proper version number.


resources/user-agents/browsers/uc-browser/uc-web-12.json
```diff
@@ -0,0 +1,130 @@
+{
```
jaydiablo: This is just a copy of one of the useragents from uc-web-11.json (the Blink based one, as I didn't see any U3 UAs on version 12.0).

PR #1809, opened 2018-02-14
[Code] Allow setting an output divison & section type on `CustomFilter` and `CustomWriterFactory`

alexander255:
I use this to do custom builds of BrowsCap JSON cache data that are shipped as part of a browser extension developed by me (size matters in this). If wanted I can also add another patch that implements the other filters as special configurations of the custom filter (reducing code concept duplication).

There is also another patch waiting to be made for allowing setting custom cache item prefix length, but this is the more important one.

### DIFF
--- a/src/Browscap/Filter/CustomFilter.php
+++ b/src/Browscap/Filter/CustomFilter.php
@@ -17,6 +17,11 @@ class CustomFilter implements FilterInterface
     private $fields = [];
 
     /**
+     * @var string
+     */
+    private $outputType = FilterInterface::TYPE_FULL;
+
+    /**
      * @var \Browscap\Data\PropertyHolder
      */
     private $propertyHolder;
@@ -24,10 +29,12 @@ class CustomFilter implements FilterInterface
     /**
      * @param PropertyHolder $propertyHolder
      * @param array          $fields
+     * @param string         $outputType
      */
-    public function __construct(PropertyHolder $propertyHolder, array $fields)
+    public function __construct(PropertyHolder $propertyHolder, array $fields, $outputType = FilterInterface::TYPE_FULL)
     {
         $this->fields         = $fields;
+        $this->outputType     = $outputType;
         $this->propertyHolder = $propertyHolder;
     }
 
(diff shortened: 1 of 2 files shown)


mimmi20:
Please change your PR:

* Don't change the Factory and the Filter. You should compose your writer by your own, if you really only write Json files (or add a new factory).
* Add a FilterCollection which implements the FilterInterface where you may add multiple Filters. Please implement it similar to the WriterCollection.
* Add tests for all things you change.

alexander255:
Thanks for your feedback!

> Don't change the Factory and the Filter. You should compose your writer by your own, if you really only write Json files (or add a new factory).

I'm actually writing PHP INI files then convert them to a JSON *cache* using a patched version of your `browser-json-generator` tool.

But maybe I should elaborte some more on what I'm actually trying to do:
Currently the `browscap-json-generator` tool generates a cache containing *all* the properties known to Browscap… which is a lot of data and too much for my use-case. I therefor have several quick-and-dirty patches in different parts of the codebase in place that reduce that extra size to acceptable levels. These I'm now trying to replace with something upstreamable / reusable for others. My idea for this was to extend the `browscap-json-generator` tool with some command-line flags that allow setting which devisions & sections (`full`, `standard`, `lite`) to add to the output, as well specifing a list of properties that shouldn't be emitted. Is that something that – in pricipal – sounds mergable to you? If yes, what would be the best approach (from your point-of-view) to allow for something like this?

> Add a FilterCollection which implements the FilterInterface where you may add multiple Filters. Please implement it similar to the WriterCollection.

Will do. Thanks for your help!

src/Browscap/Filter/CustomFilter.php
```diff
@@ -16,6 +16,11 @@ class CustomFilter implements FilterInterface
      */
     private $fields = [];
 
+    /**
+     * @var string
```
mimmi20: Please revert all changes in this filter.

Add a FilterCollection which implements the FilterInterface where you may add multiple Filters. Please implement it similar to the WriterCollection.

src/Browscap/Writer/Factory/CustomWriterFactory.php
```diff
@@ -4,6 +4,7 @@
 
 use Browscap\Data\PropertyHolder;
 use Browscap\Filter\CustomFilter;
+use Browscap\Filter\FilterInterface;
```
mimmi20: Please revert all change in this factory. Add a new one if nessesary.

asgrim:
Closing this due to inactivity. Please do re-open if you'd like to work on this again! Thank you.

PR #1824, opened 2018-02-22
Adding Samsung Browser 1/1.5/1.6

### LINKED ISSUE #1823: Add Samsung Browser 1.5 and 1.6
While looking at UAs for #1778 I was curious about Chrome versions less than `30.0.0.0` (which seems to be the first version used for the Chrome Webview), this lead me to this particular gist:

https://gist.github.com/NielsLeenheer/4daa6a9ce7f4a0f4733d

And this archived version of the Samsung developer site: http://web.archive.org/web/20150509064750/http://developer.samsung.com/technical-doc/view.do?v=T000000203L

Most importantly this bit:

>Previous :
UA String used in devices before 2015
Current :
New UA String used in
>- Mobile : 2015 and later released devices, and Android 5.0 Lollipop updated devices (with the small number of exceptions)
>
>Android Mobile Samsung Browser
Previous (Samsung Galaxy Note Edge):
>>Mozilla/5.0 (Linux; Android 4.4.4; en-au; SAMSUNG SM-N915G Build/KTU84P) AppleWebKit/537.36 (KTHML, like Gecko) Version/2.0 Chrome/34.0.1847.76 Mobile Safari/537.36
>
>Current (Samsung Galaxy S6 Edge):
>>Mozilla/5.0 (Linux; Android 5.0.2; SAMSUNG SM-G925F Build/LRX22G) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/3.0 Chrome/38.0.2125.102 Mobile Safari/537.36

It seems that the `Android WebView` UAs that have a certain Version string mixed with a particular Chrome version and are on a Samsung device are most likely an early version of the Samsung Internet Browser (versions 1.5 and 1.6). I've also noticed that there are 2.0 and 2.1 versions like this as well (before they changed the UA to use `SamsungBrowser`).

I'm planning to make changes to move most/all of the Samsung devices from Android Webview 1.5/1.6/2.0/2.1 to a new Samsung Browser file for these versions if they meet the requirements.

jaydiablo:
Fixes #1823

As mentioned in the linked issue, when researching some Android WebView UAs, I realized that most of the 1.0/1.5/1.6 UAs with Samsung devices are actually an older version of the Samsung Internet Browser (before they started labelling it as “SamsungBrowser” in the UA).  This is based on Samsung’s developer site and the Gist posted in #1823.

In these new patterns I’ve locked the Chrome version with the browser version so that any other combinations of this will be picked up by the Android Webview browser (I did look for other cases of this, and did find some, but when looking in our logs, the sources of those UAs appear to all be Data Center ISPs, which implies that they’re automated and probably spoofed).

I removed all of the Samsung devices from the 1.0/1.5/1.6 Android WebView files, mainly to see if there were any UAs in the tests that wouldn’t get picked up by my changes.  There were none (there’s one exception, which I’ll point out in some inline comments).

I added tests where possible/necessary. I also ran these changes vs. `master` and compared the results. Except for the handful of spoofed UAs I mentioned above, all UAs parse “better” with these changes than they do in `master` (more devices detected in these changes, `general Tablet` detected in some cases vs. `general Mobile Phone` (which looked like they probably are tablets)).

I have some inline comments to add to address some specific changes I made.


resources/user-agents/browsers/android-webview/android-webview-1-5-2-1.json
```diff
@@ -812,15 +617,6 @@
             "Android_4_2"
           ]
         },
-        {
-          "match": "Mozilla/5.0 (#PLATFORM##DEVICE# Build/*) applewebkit* (*kthml*like*gecko*)*Version/#MAJORVER#.#MINORVER#*Chrome*Safari*",
```
jaydiablo: This one I ended up changing in the tests so that the Samsung Browser changes pick it up. I don't believe this is a real UA, as I found the UA used on this page: http://developer.samsung.com/technical-doc/view.do?v=T000000203

>Mozilla/5.0 (Linux; Android 4.4.4; en-au; SAMSUNG SM-N915G Build/KTU84P) AppleWebKit/537.36 (KTHML, like Gecko) Version/2.0 Chrome/34.0.1847.76 Mobile Safari/537.36

I assume it's a type-o on Samsung's page. I searched our logs for UAs with this format (KTHML instead of KHTML) and the only UA I found over the past year was the one above (assuming it got picked up by bots/traffic generators).

resources/user-agents/browsers/android-webview/android-webview-generic.json
```diff
@@ -15,13 +15,6 @@
         "Comment": "Android WebView Generic"
       },
       "children": [
-        {
-          "match": "Mozilla/5.0 (#PLATFORM# Android/* Release/* Browser/applewebkit* Build/*) applewebkit* (*khtml*like*gecko*)*Version/*Chrome*Safari*",
```
jaydiablo: The only cases of this style UA I found were with Samsung devices, so I moved this pattern (but without a lot of the junk in the middle) to the Samsung Browser files as a WebKit version of the browser.

I did find some mention on wikipedia about Samsung using a WebKit version of the browser, and then moving to Chrome (https://en.wikipedia.org/wiki/Samsung_Internet_for_Android).

I looked through browscap's commit history to see when these patterns were first added, but there's no mention of why they were added, and no UAs hit them until a coverage add I did (#1331).

resources/user-agents/browsers/samsung/samsung-android-browser-3-0-to-6-0.json
```diff
@@ -215,19 +221,6 @@
             "Android_4_4"
           ]
         },
-        {
-          "match": "Mozilla/5.0 (#PLATFORM##DEVICE# Build/*) applewebkit* (*khtml*like*gecko*) SamsungBrowser/#MAJORVER#.#MINORVER#*Chrome*Safari*",
-          "devices": {
-            "SM-G388F": "Samsung SM-G388F",
```
jaydiablo: I removed devices that weren't covered by tests, with the assumption that they don't support this version of the browser (2.0 and 2.1 were removed from this file, so perhaps these devices weren't upgraded to 3.0+ on this version of Android)

tests/issues/issue-1419.php
```diff
@@ -302,19 +302,6 @@
         'standard' => true,
         'full' => true,
     ],
-    'issue-1419-G (lite)' => [
```
jaydiablo: I removed any lite or standard tests that were still detecting as "Android Webview".  The device-less patterns I added to the 1.0/1.5/1.6 require that "samsung" be present in the device, which doesn't happen for all UAs, so if the device isn't available in the Lite or Standard file, it won't detect them as Samsung Browser, but I've tried to not have mis-matching browser names in tests.

mimmi20 (approved):

PR #1826, opened 2018-02-25
auto-generate some properties

### LINKED ISSUE #1433: Remove deprecated properties from tests?

mimmi20:
With this PR the following properties are removed from the source files and genererated during the build process

- Browser_Type
- Crawler
- isSyndicationReader
- Device_Type
- isTablet
- isMobileDevice

Additional to that in the source files the strings "true" and "false" were changed to the boolean values true or false, some integer properties (e.g. Browser_Bits, CssVersion) have been converted to int.

The Models and Tests have been updated for this. Some missing tests were added too. 

This PR should fix #1433.


src/Browscap/Data/Browser.php
```diff
@@ -20,19 +25,29 @@ class Browser
...
-     * @param bool     $standard
+     * @param array  $properties
+     * @param string $type
+     * @param bool   $isLite
+     * @param bool   $standard
      */
-    public function __construct(array $properties, bool $isLite, bool $standard)
+    public function __construct(array $properties, string $type, bool $isLite, bool $standard)
     {
+        $this->type       = $type;
```
asgrim: What values can `type` be?

src/Browscap/Data/Device.php
```diff
@@ -18,17 +23,27 @@ class Device
...
-     * @param string[] $properties
-     * @param bool     $standard
+     * @param array  $properties
+     * @param string $type
+     * @param bool   $standard
      */
-    public function __construct(array $properties, bool $standard)
+    public function __construct(array $properties, string $type, bool $standard)
     {
+        $this->type       = $type;
```
asgrim: Same - do no validation done here to check `$type` is valid

src/Browscap/Data/Browser.php
mimmi20: Do want just a list of these types, or do you want a proper validation?

mimmi20: This object is a simple data container. The validation has to be setup in the factory.

asgrim: Data should be validated as it enters the place of usage (i.e. the class), rather than assume it is validated somewhere else; that's why value objects are useful; encapsulate validation of values in a value object, make it immutable so it can never accidentally become invalid (if you want to change the value, you create a new object always), and there is implicit trust that the value object is in a consistent, valid and invariant state.
Specifically in this case, it was a question really: is `$type` really an enumeration? Can the value be validated?

mimmi20: I added the validation for the `type` property in the factory, like all other validation we do.

The type property is not really an enum. The possible values are limited by the external libraries I created and used here and by the validation `browscap-php` does.

mimmi20:
@asgrim @jaydiablo Could you review this again?

tests/BrowscapTest/Data/Factory/DeviceFactoryTest.php
```diff
@@ -33,12 +33,40 @@ public function testBuildWithoutStandardProperty() : void
         $this->object->build($deviceData, $deviceName);
     }
 
+    /**
+     * @throws \Assert\AssertionFailedException
+     */
+    public function testBuildWithWrongDeviceType() : void
+    {
+        $this->expectException(\UnexpectedValueException::class);
```
asgrim: Just a minor nitpick, but ideally `expectException*` should be just before where the exception happens, rather than at the start of the method. No need to change anything here, but just maybe more logical :)

asgrim:
Thanks @mimmi20 :+1:

mimmi20:
Thanks

PR #1839, opened 2018-03-25
remove duplicate definitions

### LINKED ISSUE #1820: Adding a device twice does not cause an error
If a device is added in different files, the result depends on the device which is added first.

To avoid this, this should cause an error.

mimmi20:
I added a check for duplicate definitions in the browsers, devices and useragent files. I removed the duplicates found by this check.

Fixes #1820.


mimmi20:
@asgrim Do you have time to review this PR?

PR #1845, opened 2018-04-02
Adding some missing browser/platform versions from 2018-02/03

### LINKED ISSUE #1844: Missing Browser & Platform versions from 2018-02 and 2018-03
I wasn't able to get around to doing this for Feb, so doing both months with this update.

iOS 11.3 is out, so will be adding that to the platforms list and updating different browsers/apps that need version or platform bumps.

jaydiablo:
This fixes #1844.

I pulled a bunch of UAs from February/March that were missing the browser and/or platform version but shouldn’t have been.

Most of the changes are just browser version updates for these UAs.  Though there are a number of new versions of iOS added (particularly iOS 11.3 which was recently released).

I did tweak the headless chrome pattern a tiny bit to pick up some UAs I was seeing that had something like this `HeadlessChrome/65.0.x.x.x` which the current pattern wasn’t catching.

Also noticed some PaleMoon 26.x UAs that weren’t getting picked up by the versioned file due to the generic file having a longer matching pattern.


resources/user-agents/browsers/palemoon/palemoon-20-25.json
```diff
@@ -1,6 +1,6 @@
 {
   "division": "PaleMoon #MAJORVER#.#MINORVER#",
-  "versions": ["25.9", "25.8", "25.7", "25.6", "25.5", "25.4", "25.3", "25.2", "25.1", 25, "24.9", "24.8", "24.7", "24.6", "24.5", "24.4", "24.3", "24.2", "24.1", 24, "20.3", "20.2", "20.1", 20],
```
jaydiablo: I did remove some versions to make this better align with what is listed in PaleMoon's FTP archive (ftp://archive.palemoon.org/) the only one that I left that isn't in their FTP is 24.9, which browscap has a test for.  I did search a bit to see if it was actually a release that didn't make it to FTP for some reason, but nothing concrete came up. There was some discussion about PaleMoon 25.1 acting as Firefox 24.9, but that seemed related to addons.mozilla.org and not really the useragent.  I left the test and the version anyway.

resources/user-agents/browsers/uc-browser/uc-web-12.json
```diff
@@ -126,13 +126,23 @@
...
+            "Android_8_1", "Android_8_0",
             "Android_7_1", "Android_7_0",
             "Android_6_0",
             "Android_5_1", "Android_5_0",
             "Android_4_4", "Android_4_3", "Android_4_2", "Android_4_1", "Android_4_0",
             "Android"
           ]
+        },
+        {
+          "match": "Mozilla/5.0 (*Windows *) applewebkit* (*khtml*like*gecko*) Chrome/??.* Safari/* UCBrowser/#MAJORVER#.#MINORVER#*",
```
jaydiablo: This is just copied from the 11.x file.

jaydiablo:
Oops, forgot to run standard tests, will fix.

PR #1851, opened 2018-04-07
Aligning IE to supported versions of windows, adding a lot of tests for coverage

### LINKED ISSUE #1846: IE and supported versions of Windows
Was looking at this a bit today, and it seems that browscap has more Windows versions defined for different versions of IE than are supported by the operating systems themselves. For example, I was testing loading IE 9.0 on Windows XP today, which isn't possible (not possible to download from MS, if you can get an EXE from another source, it doesn't install on XP).

I looked around a bit for the official support of IE across different versions of Windows, and found this doc: https://support.microsoft.com/en-us/help/11531/internet-explorer-system-requirements

I also googled around a bit for some other cases (like is it possible to install anything > IE 6.0 on Windows 2000) and found various support articles with answers from MS reps (or community moderators) that suggest it isn't possible to mix IE X.0 with Windows X. Ultimately I came to this conclusion:

## IE 6.0
This isn't in that linked document, but using wikipedia article (https://en.wikipedia.org/wiki/Internet_Explorer_6) and other searches on the web, this is what I've come up with:
 - Windows XP and lower only (Win98, NT, Win2000, WinME)
 - **Maybe** Windows 7 in XP Mode (I haven't tried to confirm this yet, XP mode may show itself as XP in the UA, but maybe it shows Windows 7)

### to-do
 - [ ] Remove Windows 8
 - [ ] Remove Vista
 - [ ] Add Windows 98
 - [x] Test Windows 7
 - [ ] Remove Windows 7

## IE 7.0
Also not in that MS document, but wikipedia (https://en.wikipedia.org/wiki/Internet_Explorer_7) and related searches.
 - Windows XP and Windows Vista only
 - **Maybe** Windows 7 in XP Mode (same considerations as IE 6.0)

### to-do
 - [ ] Remove Windows 8
 - [ ] Remove Windows 2000
 - [x] Test Windows 7
 - [ ] Remove Windows 7

## IE 8.0
 - Windows XP
 - Windows Vista
 - Windows 7

### to-do
 - [ ] Remove Windows 8

## IE 9.0
 - Windows Vista
 - Windows 7

### to-do
 - [ ] Remove Windows 8
 - [ ] Remove Windows XP

## IE 10.0
 - Windows 7
 - Windows 8

### to-do
 - [ ] Remove Windows XP
 - [ ] Remove Windows 8.1
 - [ ] Remove Windows 10

## IE 11.0
 - Windows 7
 - Windows 8.1 (yes, just 8.1, not 8.0. Apparently not possible to install IE 11 on 8.0 systems)
 - Windows 10

### to-do
 - [ ] Remove Windows XP
 - [ ] Remove Windows 8
 - [ ] Remove Windows Vista

I'll make these changes and see what tests fail and report back.

jaydiablo:
Fixes #1846.

I gathered a bunch of UAs that browscap detected as IE from the past year and made tests from those before removing the invalid Windows versions from the patterns.

Any tests that failed I moved to the invalid-versions file and changed their platform to either `Win32` or `Win64` based on how they were detected. Any tests that passed I left in for coverage.

I also removed any patterns that had 0 coverage after adding all of these tests.

The addition of the new `*Windows*WOW64*` platform was to make sure that Compatibility mode is still detected as the proper browser when the platform is bad, and not the emulated browser. I only added those two *64 Windows platforms to the patterns that are Compatibility mode.

I compared these changes to `master` for a bunch of real-world UAs that I’ve collected over the past few months and made sure that valid UAs are still detected properly (I did this in chunks of 50,000 UAs for each month for the past few months, and in each case, only about 20 - 40 UAs detected differntly. Most of them being IE 11.0 on Windows 8 (which as far as I can tell, isn’t possible, as IE 11.0 was only available in the Windows 8.1 update)).


resources/user-agents/apps/itunes/itunes-1-2-to-4-7.json
```diff
@@ -31,7 +31,7 @@
           "match": "iTunes/#MAJORVER#.#MINORVER#* (#PLATFORM#)*",
           "device": "Windows Desktop",
           "platforms": [
-            "Windows_64",
+            "Windows_x64",
```
jaydiablo: These were the only UAs that used the "Windows_64" platform, I just changed its name to better match the `WOW64` vs `x64` platforms.

PR #1885, opened 2018-06-09
add MailPoet Cron Service

### LINKED ISSUE #1876: MailPoet Cronjob User Agent Not Recognized
On running a lookup, Browscap didn't appear to catch the User Agent that MailPoet uses when triggering cronjob urls. This may or may not be something necessary to consider, as it's fairly custom; but I did note many crawlers have been included as part of Browscap.

MailPoet seems to use the following User Agent string:
`Mozilla/5.0 (compatible; MailPoet Cron/1.0; +http://docs.mailpoet.com/article/241-cron-service)`

https://www.mailpoet.com/
http://docs.mailpoet.com/article/241-cron-service

mimmi20:
Fixes #1876.


asgrim:
Thank you!

PR #1887, opened 2018-06-11
Aborting useragent test suites when INI file building throws an exception

### LINKED ISSUE #1333: If the INI file can't be built, useragent tests should abort?
I'm not sure if this is possible within the way that PHPUnit operates without using something like `die` or `exit` but it would be nice if when the INI build step of the tests fails, that the useragent tests would abort at that point (and output the error message that the INI builder dumps).

There are times where I mess up a JSON file when making a change, kick off tests, switch windows, come back and see that each test is just erroring out.  Usually I then run the `build` command manually to see why the test failed (if the JSON error isn't obvious in my editor).

jaydiablo:
Fixes #1333.

I ran into this again today, so figured I'd try to put together a solution for it.

I referenced an undefined platform in an update I'm doing, and this was the output when I ran the tests:

```
PHPUnit 7.2.4 by Sebastian Bergmann and contributors.

EEEEEEEEEEEEEEEEEEEEEEEEEEEEEE
(truncated)
```

Sometimes I kick off the tests and then come back to see how they did (I was watching them in this case), which is annoying when it's just a screen full of `E`s.  It's also not possible to know why the tests are failing here without letting the suite finish (which actually takes longer than a successful run), or aborting the suite and either running the build manually, or using the `--stop-on-failure` flag to phpunit.

With these changes, this error will get caught during the INI building process and then output will look something like this:

```
PHPUnit 7.2.4 by Sebastian Bergmann and contributors.

Browscap ini file could not be built in UserAgentsTest\V3\FullTest test class, there was an uncaught exception: OutOfBoundsException (Platform "iOS_I_11_3" does not exist in data)
```

IMO, this is more helpful, because the tests will abort immediately and will output a helpful error message so I can correct the issue and re-run the suite.

I don't know of a "clean" way to abort phpunit execution during the `setUpBeforeClass`, so used `die` here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ cache:
 stages:
   - composer validate
   - validate sources and tests
+  - style check
   - static code analysis
   - test
   - test with coverage
@@ -54,7 +55,7 @@ jobs:
       install: skip
       script: composer validate --strict
 
-    - stage: static code analysis
+    - stage: style check
       php: 7.1
       script: php -n -d memory_limit=768M vendor/bin/php-cs-fixer fix --dry-run -vv
 
--- a/tests/UserAgentsTest/V3/FullTest.php
+++ b/tests/UserAgentsTest/V3/FullTest.php
@@ -73,45 +73,54 @@ class FullTest extends TestCase
 
         $version = (string) $buildNumber;
 
-        $logger               = new NullLogger();
-        $writerCollection     = new WriterCollection();
-        self::$propertyHolder = new PropertyHolder();
-        self::$filter         = new FullFilter(self::$propertyHolder);
-        self::$writer         = new IniWriter($buildFolder . '/full_php_browscap.ini', $logger);
-        $formatter            = new PhpFormatter(self::$propertyHolder);
-        self::$writer->setFormatter($formatter);
-        self::$writer->setFilter(self::$filter);
-        $writerCollection->addWriter(self::$writer);
+        try {
+            $logger               = new NullLogger();
+            $writerCollection     = new WriterCollection();
+            self::$propertyHolder = new PropertyHolder();
+            self::$filter         = new FullFilter(self::$propertyHolder);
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -54,7 +55,7 @@ jobs:
       install: skip
       script: composer validate --strict
 
-    - stage: static code analysis
+    - stage: style check
```
jaydiablo: I didn't think `php-cs-fixer` belongs in the `static code analysis` step.

tests/UserAgentsTest/V3/FullTest.php
```diff
@@ -73,45 +73,54 @@ public static function setUpBeforeClass() : void
 
         $version = (string) $buildNumber;
 
-        $logger               = new NullLogger();
```
mimmi20: Instead of these changes, could we add a new build stage, where we just call the CLI build command?

jaydiablo: These blocks of code vary slightly for V3/V4, how would that be handled in the CLI Command?

mimmi20: The code you moved into the `try` block is nearly the same, as in the CLI build command.

Instead of changing 6 files, I would prefer to add a new step in the build pipeline (travis), where we just run the CLI build command.

jaydiablo: Well, this isn't just for Travis runs, I'd make use of it when running tests locally as well.

jaydiablo: If you think it makes sense, I could try to call the build command and remove a lot of the duplicated code between the three (or 6, though this seems trickier to do with a single build command unless it branches a bit based on the installed browscap version) test case files.

I know it's possible to call symfony commands from other commands, but not sure how simple it is from non-console code.

PR #2007, opened 2018-11-13
partly replace jsonlint

mimmi20:
After I learned that JsonLint is slow, I tried to replace it.

This PR does:
1. replace JsonLint outside the validate functions
2. extract some duplicate code into command helpers
3. cleanup unused fixture files
4. allow to sort the data inside the resource files


mimmi20:
@asgrim @jaydiablo Could you review this?

PR #2076, opened 2019-03-24
add Amazon Echo Show and Echo Dot

### LINKED ISSUE #2074: Amazon Echo user agents
see sanchezzzhak/node-device-detector#8 for more details

mimmi20:
Fixes #2074.


resources/devices/mobile-device/amazon.json
```diff
@@ -0,0 +1,24 @@
+{
+  "Amazon AEOKN": {
+    "type": "mobile-device",
```
jaydiablo: Is Mobile Device the proper device type for these sorts of devices?

I'm not sure what *is* the right device type, they just don't seem very mobile to me.

mimmi20: Because `browscap-php` validates the device types, its not possible to use something like `speaker`.

mimmi20: This is the list of possible types:
'car-entertainment-system', 'console', 'desktop', 'digital-camera', 'ebook-reader', 'feature-phone', 'fone-pad', 'mobile-console', 'mobile-device', 'mobile-phone', 'smartphone', 'tablet', 'tv', 'tv-console', 'unknown'

jaydiablo: Right now `console` is only being used for game consoles?  Like Playstation/Switch/Wii/Xbox?

Should a new device type be added for these things? I'm not sure what else they might fit in, `console` seems like the most appropriate I suppose, as it's generic enough to not rule out.

They're not really mobile (have to be plugged in to a wall to function), certainly not a phone or tablet (rarely have a screen, though the Echo Show has a video screen I think).

I think `console` or something new makes the most sense here.  I'm curious what other parsers do with them (if they detect them at all). UA Parser, Matomo, WhichBrowser may since they're usually pretty comprehensive.

jaydiablo: Eh, not much help there.

Matomo detected it as a generic Tablet (didn't detect the device brand/name).
UA Parser has it as a generic android device (no type provided).
WhichBrowser has it as a Tablet as well (didn't detect the device brand/name).

That was on this UA:

```
Mozilla/5.0 (Linux; Android 5.1.1; AEOKN Build/LVY48F; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/59.0.3071.125 Safari/537.36
```

jaydiablo: The Show may be best as a Tablet though, now that I'm looking at the product:

https://www.amazon.com/All-new-Echo-Show-2nd-Gen/dp/B077SXWSRP

But the Dot wouldn't be classified as such.

Those three parsers don't detect the Dot either. Matomo and UA Parser don't detect the device at all on this UA:

```
AndroidDownloadManager/5.1.1 (Linux; U; Android 5.1.1; AEOBC Build/LVY48F)
```

Which browser detects it as a tablet (generic).

So maybe the Show as a Tablet and the Dot as a "Console"?  Thoughts?

PR #2164, opened 2019-10-02
Fix 2000 generation date option

### LINKED ISSUE #2000: Use git tag date for `GJK_Browscap_Version`.`Released` date
Currently we just use the current date/time for making the `Released` date:

```ini
[GJK_Browscap_Version]
Version=6000030
Released=Fri, 26 Oct 2018 12:54:15 +0100
```

Ideally we should use the date of the git tag for `6.0.30` so the date is always consistent, e.g.:

```ini
[GJK_Browscap_Version]
Version=6000030
Released=Tue Aug 14 09:03:56 2018 +0000
```

Also the date should be adjusted for UTC.

asgrim:
Fixes #2000 

Allows setting generation date externally (optional).

### DIFF
--- a/src/Browscap/Generator/BuildGenerator.php
+++ b/src/Browscap/Generator/BuildGenerator.php
@@ -65,8 +65,8 @@ final class BuildGenerator implements GeneratorInterface
      * Entry point for generating builds for a specified version
      *
      * @param string            $buildVersion
-     * @param bool              $createZipFile
      * @param DateTimeImmutable $generationDate
+     * @param bool              $createZipFile
      *
      * @throws \Exception
      * @throws \Assert\AssertionFailedException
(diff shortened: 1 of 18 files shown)


src/Browscap/Generator/BuildGenerator.php
```diff
@@ -63,19 +64,21 @@ public function __construct(
     /**
      * Entry point for generating builds for a specified version
      *
-     * @param string $buildVersion
-     * @param bool   $createZipFile
+     * @param string            $buildVersion
+     * @param bool              $createZipFile
+     * @param DateTimeImmutable $generationDate
```
mimmi20: This param is on the wrong position

tests/BrowscapTest/Data/Factory/DataCollectionFactoryTest.php
```diff
@@ -62,7 +63,7 @@ public function testCreateDataCollection() : void
         $logger = $this->createMock(Logger::class);
 
         $collection = $this->getMockBuilder(DataCollection::class)
-            ->setConstructorArgs([$logger])
+            ->setConstructorArgs([$logger, new DateTimeImmutable()])
```
mimmi20: The Datetime is not a Contructor value.

asgrim:
Thanks for review @mimmi20 !

PR #2207, opened 2020-01-20
add new version of Model S Browser

### LINKED ISSUE #2205: Tesla Car Browser detected incorrectly
see matomo-org/device-detector#6164 for more details

mimmi20:
Fixes #2205.


resources/browsers/browser.json
```diff
@@ -1548,7 +1548,7 @@
     "lite": false,
     "standard": true
   },
-  "model s browser": {
+  "model-s-browser": {
     "type": "browser",
     "properties": {
       "Browser": "Model S Browser",
```
jaydiablo: Tesla makes more than one model of car, should this be changed to be "Tesla Browser" or something like that (there is a different browser called the Tesla Browser (like Nikola Tesla, not the car company) that may now be defunct (https://web.archive.org/web/20180617181316/https://www.teslabrowser.com/ru), but something to consider)?

asgrim: I think to test this effectively, someone should buy us each a Tesla :grin:

PR #2230, opened 2020-02-18
add new versions, remove minor versions, add new tests

mimmi20:
Fixes #2213.
Fixes #2220.
Fixes #2224.
Fixes #2227.

PR #2244, opened 2020-03-09
Add new Chrome versions (82, 81, 80)

kondurake:


mimmi20:
added versions are also added in PR #2230.

PR #2245, opened 2020-03-09
Ability to specify a version range instead of specifying each version

kondurake:
For example, instead of:
```
{
    "versions": [88, 73, 72, 71, 70, 69, 68, 67, 66, 65, 64, 63, 62, 61, 60, 59, 58, 57, 56, 55, 54, 53, 44]
}
```

We can use:
```
{
    "versions": [88, {"start":53,"end":73}, 44]
}
```

### DIFF
--- a/src/Browscap/Data/Division.php
+++ b/src/Browscap/Data/Division.php
@@ -65,8 +65,17 @@ class Division
         $this->userAgents = $userAgents;
         $this->lite       = $lite;
         $this->standard   = $standard;
-        $this->versions   = $versions;
         $this->fileName   = $fileName;
+
+        foreach ($versions as $version) {
+            if (isset($version['start']) && isset($version['end'])) {
+                foreach (range($version['start'], $version['end']) as $version) {
+                    $this->versions[] = $version;
+                }
+            } else {
+                $this->versions[] = $version;
+            }
+        }
     }
 
     public function isLite() : bool


asgrim:
I like this addition! Seems like we could use this a lot. If you're willing, could you take a look at the failing build, and possibly add some tests for this please? If not one of us will get to this eventually 😁 great idea, thank you, let's work on this 👍

src/Browscap/Data/Division.php
```diff
@@ -65,8 +65,17 @@ public function __construct(
         $this->userAgents = $userAgents;
         $this->lite       = $lite;
         $this->standard   = $standard;
-        $this->versions   = $versions;
         $this->fileName   = $fileName;
+
+        foreach ($versions as $version) {
```
mimmi20: We have a Json Schema for Division files in `/schema/divisions.json` to validate our resource files.
Please search for the `versions` property and replace it with this:

```json
"versions": {
      "type": "array",
      "items": {
        "anyOf": [
          {
            "type": "string"
          },
          {
            "type": "integer"
          },
          {
            "type": "object",
            "properties": {
              "start": {
                "type": "integer"
              },
              "end": {
                "type": "integer"
              }
            },
            "required": [
              "start",
              "end"
            ]
          }
        ]
      }
    }
```

kondurake:
@asgrim I'm very glad you like it. Before submitting the patch I generated some builds with `bin/browscap build` and the entire range versions of user agents have been here. This was the only thing I checked.

Well, I can try to add some tests, but I'm new here. This is why I think tests should be written by experienced contributors.

PR #2336, opened 2020-08-03
Add automatic releases workflow

asgrim:
Adds workflow from https://github.com/laminas/automatic-releases/ to automatically release.

NOTE: the default branch has now changed to `6.0.x` instead of `master`.

### DIFF
--- a/.github/workflows/release-on-milestone-closed.yml
+++ b/.github/workflows/release-on-milestone-closed.yml
@@ -1,0 +1,47 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+name: "Automatic Releases"
+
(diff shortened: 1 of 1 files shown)


mimmi20:
> NOTE: the default branch has now changed to `6.0.x` instead of `master`.

Does that mean, that for each new major release a new branch is required?

asgrim:
> > NOTE: the default branch has now changed to `6.0.x` instead of `master`.
> 
> Does that mean, that for each new major release a new branch is required?

Yes but even further, a new branch for each *minor* release is used, but the automation will do this for us, and update the default branch each time.

mimmi20 (approved):

PR #2404, opened 2020-12-07
Update auto releases action

asgrim:
- Don't create merge-up PR
 - Don't auto-create and switch release branches (we only bump patch version usually!)
 - Create milestones though, useful
 - Use org token, just in case we ever want to trigger a release workflow in future

### DIFF
--- a/.github/workflows/release-on-milestone-closed.yml
+++ b/.github/workflows/release-on-milestone-closed.yml
@@ -25,23 +25,3 @@ jobs:
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
-
-      - name: "Create Merge-Up Pull Request"
-        uses: "laminas/automatic-releases@1.0.1"
-        with:
-          command-name: "laminas:automatic-releases:create-merge-up-pull-request"
-        env:
-          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
-          "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
-          "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
-          "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
-
-      - name: "Create and/or Switch to new Release Branch"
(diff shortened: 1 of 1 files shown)


.github/workflows/release-on-milestone-closed.yml
```diff
@@ -21,27 +21,17 @@ jobs:
...
-          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
 
-      - name: "Create Merge-Up Pull Request"
-        uses: "laminas/automatic-releases@1.0.1"
+      - name: "Create new milestones"
+        uses: "laminas/automatic-releases@v1"
```
mimmi20, line 30: The version should be the same as in line 20

mimmi20, line 30: If the following is added to the dependabot config, dependabot will also update the github actions:

```yaml

  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: daily
      time: '04:00'
    open-pull-requests-limit: 10
    reviewers:
      - ...
    assignees:
      - ...
    labels:
      - "dependencies"
    commit-message:
      include: "scope"
      prefix: "github-actions"
```

mimmi20 (changes requested):

mimmi20:
If the following is added to the dependabot config, dependabot will also update the github actions:

```yaml

  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: daily
      time: '04:00'
    open-pull-requests-limit: 10
    reviewers:
      - ...
    assignees:
      - ...
    labels:
      - "dependencies"
    commit-message:
      include: "scope"
      prefix: "github-actions"
```

.github/workflows/release-on-milestone-closed.yml
asgrim: Good call on the inconsistency; I've updated them both to be `@v1` though, which should negate the need to have dependabot update things I think?

mimmi20: dependabot would also update to v2.

mimmi20 (approved):

.github/workflows/release-on-milestone-closed.yml
asgrim: Updates to major versions (assuming semver) may have breaking changes, let's keep it to just v1 for now I think :)

PR #2421, opened 2021-01-11
update dependencies and tests

mimmi20:
This PR wants:
- update dependencies to actual versions (for the supported PHP versions)
- remove some dependencies which were not used, or were replaced by using docker
- reformat `dependabot.yml`, add a section to update Github actions
- remove deprecated composer update option `--no-suggest`
- update coding style definition, disable the `after_heredoc` options as this will crash for PHP < 7.3
- fix coding style issues and issues found by phpstan
- update tests


.github/dependabot.yml
```diff
@@ -1,42 +1,62 @@
...
-  - dependency-name: symfony/finder
-    versions:
-    - ">= 0"
-  - dependency-name: twig/twig
-    versions:
-    - ">= 0"
-  commit-message:
-    prefix: Build
-    include: scope
+  - package-ecosystem: composer
```
asgrim, line 6: yikes, were the line endings `\r\n` before? O_O

.php_cs
```diff
@@ -23,8 +23,13 @@ return PhpCsFixer\Config::create()
         '@PHP71Migration' => true,
         '@PHP71Migration:risky' => true,
         '@PHP73Migration' => true,
+        '@PHP74Migration' => true,
```
asgrim, line 26: All this CS maintenance things we should switch to `doctrine/coding-standard`, so much simpler than maintaining this ruleset I think .... thoughts?

asgrim (approved):
Hmm is there anything we can do to stop codecov "failing" the build? I think it's just because coverage dropped or something, IDK if that should be a build-fail reason?

.php_cs
mimmi20: All 3 projects (`browscap`, `browscap-php` and `browscap-site`) have their own coding styles and tooling sets. If we change the coding standard, we should unify the coding standard and the tools over all projects (in a seperate PR).

asgrim: I am ok with that; `doctrine/coding-standard` is easy to pull in & apply (`vendor/bin/phpcs` to check, `vendor/bin/phpcbf` to update).

mimmi20:
> Hmm is there anything we can do to stop codecov "failing" the build? I think it's just because coverage dropped or something, IDK if that should be a build-fail reason?

I dont think that this is an issue, as long we dont define these tests as `required`.

asgrim:
> I dont think that this is an issue, as long we dont define these tests as `required`.

Fair enough :) just on glance, all I see is "failed build" so I tend to ignore it heh

Issue #2479, opened 2021-04-28
Safari 14.1 released


rajivraman:
Safari 14.1 was released on April 26 and it appears to be missing from the latest "lite" browscap file.

PR #2517, opened 2021-08-20
add safari 14.1

guetj:
#2479
A small contribution in an attempt to add the support of safari 14.1.
I think this is ok: unit test have been added, build completes successfully and the browscap.zip generated seems to include the definition of Safari 14.1.
However I recommend to be cautious with this pull request: I am totally new to browscap, so I may be missing something.


asgrim (approved):

guetj:
I see that there is a failure on the CI build. I will need help to fix this.

mimmi20 (approved):

PR #2543, opened 2021-12-03
update dependencies

mimmi20:


.github/workflows/continuous-integration.yml
```diff
@@ -6,71 +6,73 @@ on:
...
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v2.4.0"
+        with:
+          # Disabling shallow clone is recommended for improving relevancy of reporting
+          fetch-depth: 0
 
       - name: "Install PHP"
-        uses: "shivammathur/setup-php@2.16.0"
+        uses: "shivammathur/setup-php@2.15.0"
```
asgrim: Was this downgrade intentional? :thinking:

.github/workflows/continuous-integration.yml
```diff
@@ -6,71 +6,73 @@ on:
...
     strategy:
       fail-fast: false
+
       matrix:
-        php-versions:
-          - "7.1"
-          - "7.2"
-          - "7.3"
+        operating-system:
+          - "ubuntu-18.04"
```
asgrim, line 66: Do we need to test on such an old `ubuntu` still? 20.04 is the latest LTS and is sufficient IMO

.github/workflows/continuous-integration.yml
```diff
@@ -79,214 +81,242 @@ jobs:
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v2.4.0"
+        with:
+          # Disabling shallow clone is recommended for improving relevancy of reporting
+          fetch-depth: 0
 
       - name: "Install PHP"
-        uses: "shivammathur/setup-php@2.16.0"
+        uses: "shivammathur/setup-php@2.15.0"
```
asgrim: Downgrade intentional?

.github/workflows/continuous-integration.yml
```diff
@@ -79,214 +81,242 @@ jobs:
...
+          coverage: "none"
+          tools: "composer:v2"
 
       - name: "Determine composer cache directory"
         id: "determine-composer-cache-directory"
         run: "echo \"::set-output name=directory::$(composer config cache-dir)\""
 
       - name: "Cache dependencies installed with composer"
-        uses: "actions/cache@v2.1.7"
+        uses: "actions/cache@v2.1.6"
```
asgrim: Looks to be downgraded unintentionally too?

composer.json
```diff
@@ -29,37 +29,36 @@
...
+    "mimmi20/ua-browser-type": "^8.0.5",
+    "mimmi20/ua-device-type": "^8.0.5",
+    "monolog/monolog": "^2.3.5",
     "seld/jsonlint": "^1.8.3",
-    "symfony/console": "^4.4.18 || ^5.2.1",
-    "symfony/finder": "^4.4.18 || ^5.2.1",
-    "twig/twig": "^2.13.1 || ^3.2.1"
+    "symfony/console": "^v5.4.0",
+    "symfony/finder": "^v5.4.0",
+    "twig/twig": "^v3.3.4"
```
asgrim, line 46: Just noticed this - why do we need twig? :thinking:

.github/workflows/continuous-integration.yml
mimmi20: In the company I am working we are using this version, and there are currently no plans to update. To make sure, I am able to install the project here to work on it sometimes, I would like to keep the old ubuntu version for this test.

composer.json
mimmi20: We have a console command to rewrite the files in the resources directory. Twig is used to render the files to keep the structure we have.

composer.json
```diff
@@ -28,42 +27,43 @@
...
   },
   "require-dev": {
     "ext-zip": "*",
-    "browscap/browscap-php": "^4.3.0 || ^5.0.0",
+    "browscap/browscap-php": "dev-updates",
```
asgrim: Needs making stable before merge

composer.json
```diff
@@ -28,42 +27,43 @@
...
     }
   ],
+  "homepage": "https://github.com/browscap/browscap",
+  "support": {
+    "issues": "https://github.com/browscap/browscap/issues",
+    "source": "https://github.com/browscap/browscap"
+  },
   "require": {
-    "php": "^7.1.3",
+    "php": "^7.4.3",
```
asgrim: We could probably go to PHP `^8.1.0` now I think?

asgrim (commented):
Overall looks good :+1:

src/Browscap/Data/Expander.php
```diff
@@ -205,6 +211,8 @@ private function parseUserAgent(UserAgent $uaData, bool $liteUa, bool $standardU
         foreach ($uaData->getChildren() as $child) {
             $this->patternId['child'] = $i;
 
+            assert(is_array($child));
+
             if (isset($child['devices']) && is_array($child['devices'])) {
                 // Replace our device array with a single device property with our #DEVICE# token replaced
                 foreach ($child['devices'] as $deviceMatch => $deviceName) {
```
asgrim, line 218: fwiw, if we went to PHP 8+, the line L213 would need updating, since $deviceMatch is thought to be an `integer`:

```
PHP Fatal error:  Uncaught TypeError: str_replace(): Argument #2 ($replace) must be of type array|string, int given in /home/james/workspace/browscap-site/vendor/browscap/browscap/src/Browscap/Data/Expander.php:213
```

mimmi20: > fwiw, if we went to PHP 8+, the line L213 would need updating, since $deviceMatch is thought to be an `integer`

`$deviceMatch` is a string

composer.json
mimmi20: For Windows there not an installable verison with PHP 8.1 at the moment, maybe next week. I upgraded my system to PHP 8 today

src/Browscap/Data/Expander.php
mimmi20: How did you get this error? `$deviceMatch` should be an string.

composer.json
mimmi20: As soon as browscap/browscap-php#406 is released

asgrim: browscap-php 6.0.0 is now released: https://github.com/browscap/browscap-php/releases/tag/6.0.0

src/Browscap/Data/Expander.php
asgrim: I was trying `browscap-site` out with PHP 8.1 (which I'm running locally); I'm not sure why `$deviceMatch` is sometimes an `int`, I didn't manage to track it; casting to `(string) $deviceMatch` where it gets used in `str_replace` might be OK; but it would be interesting to find out why it's sometimes an `int` nonetheless, to see if the issue can be fixed further up :)

mimmi20: We have a device with this replacement `"2014818": "Xiaomi 2014818"`. It looks like json_decode converts this string into an integer.

composer.json
mimmi20: I created issue #2547 for this

src/Browscap/Data/Expander.php
asgrim: Hah! Funnily enough, this was recently discussed on the PHP internals list :grin: https://externals.io/message/116735

asgrim (approved):

asgrim:
Thanks for this @mimmi20 ! Great work

PR #2557, opened 2022-01-10
Fix phpstan + removed .travis.yml

asgrim:


.github/workflows/continuous-integration.yml
```diff
@@ -8,7 +8,9 @@ name: "Continuous Integration"
 
 env:
   PHP_EXTENSIONS: "json, zip, opcache, mbstring"
-  PHP_INI_VALUES: "opcache.enable=1, opcache.enable_cli=1, opcache.fast_shutdown=0, zend.assertions=1, assert.exception=On, zend.exception_ignore_args=0"
+  # Note - opcache.enable_cli=1 seems to be crashing phpstan, but ONLY in CI, could not reproduce locally
```
asgrim, line 11: After much push & retry in CI, eventually tracked down the `opcache.enable_cli=` INI setting was causing phpstan to crash, but I could not reproduce this locally, weirdly. Might be worth investigating in the future, but for now, this fixes the static analysis build.

mimmi20 (approved):

PR #2761, opened 2023-01-08
Allow FilterCollection to be used anywhere that FilterInterface can be used

chappy84:
What it says on the tin really.
I wanted to combine the `CustomFilter` with the `StandardFilter`, so I could reduce the number of fields, but retain the smaller size of the standard file. The `FilterCollection` couldn't be passed to `WriterCollection::setFilter` as it didn't implement `FilterInterface`, but it seemed to make sense allow a combination of filters to be passed to this, and `FilterCollection` seemingly being intended for this purpose, hence these updates.


asgrim (approved):
No BC breaks here, and seems reasonable

PR #2767, opened 2023-01-13
Allow engineers to easily mix and match the various existing filter methods to create filters, and thus files however they wish

chappy84:
e.g. a standard file type with custom fields would just need:

```php
class CustomFieldStandardFilter implements \Browscap\Filter\FilterInterface
{
    use \Browscap\Filter\Division\StandardDivisionFilter;
    use \Browscap\Filter\Property\FieldPropertyFilter;
    use \Browscap\Filter\Section\StandardSectionFilter;

    public function getType(): string
    {
        return 'STANDARD_WITH_CUSTOM_FIELDS';
    }
}
```
This reduces the current custom file size down from ~49MB to ~17MB, which obviously is very beneficial however anyone wishes to use it.
It also removes the direct link between Divisions, Sections & Properties. I understand from the data structure there's at-least a fairly strong link between division and section, however properties can quite easily be disconnected from the others outside of the normal file types produced.

Example, I had was trying to produce a standard file with the properties:
```
Parent, Cookies, Crawler, JavaScript
```
Using a `FilterCollection` with both `CustomFilter` and `StandardFilter`, the `Cookies` and `Javascript` properties would be dropped despite the `CustomFilter` being told I wanted them, as the `StandardFilter` removes them when calling `PropertyHolder::isStandardModeProperty`.

With the existing code, the engineer would have to use something like this to re-use existing code, and reduce future maintenance to a minimum:
```php
class CustomFieldStandardFilter extends \Browscap\Filter\CustomFilter implements \Browscap\Filter\FilterInterface
{
    public function __construct(
        \Browscap\Data\PropertyHolder $propertyHolder,
        array $fields,
        private \Browscap\Filter\StandardFilter $standardFilter
    ) {
        parent::__construct($propertyHolder, $fields);
    }

    public function getType(): string
    {
        return 'STANDARD_WITH_CUSTOM_FIELDS';
    }

    public function isOutput(\Browscap\Data\Division $division): bool
    {
        return $this->standardFilter->isOutput($division);
    }

    public function isOutputSection(array $section): bool
    {
        return $this->standardFilter->isOutputSection($section);
    }
}
```

The existing public interfaces to classes are still there, and as the project now requires PHP 8, traits won't break anything, so there's no BC breaks here.

It may be beneficial for this to go into 6.2.x?, but I thought I'd put it forward as a suggestion.


phpstan.neon
```diff
@@ -27,6 +27,7 @@ parameters:
     - '~Parameter #1 \$array of method Browscap\\Parser\\IniParser::sortArrayAndChildArrays\(\) expects array<array<string>\|string>, array<array<array<string>\|string>\|string> given~'
     - '~Method Browscap\\Parser\\IniParser::sortArrayAndChildArrays\(\) should return array<array<array<string>\|string>> but returns array<array<array<array<string>\|string>\|string>\|string>~'
     - '~Expression in empty\(\) is not falsy~'
+    - '~Call to an undefined method PHPUnit\\Framework\\MockObject\\MockObject::isOutput~'
```
chappy84, line 30: I added this as it's raised on the objects created using `getMockForTrait`, which is used to test traits

src/Browscap/Filter/Property/FullPropertyFilter.php
```diff
@@ -0,0 +1,29 @@
...
+use Browscap\Data\PropertyHolder;
+use Browscap\Writer\WriterInterface;
+
+/**
+ * this filter is responsible to select properties for the "full" version of the browscap files
+ */
+trait FullPropertyFilter
+{
+    /** @throws void */
+    public function __construct(private PropertyHolder $propertyHolder)
```
mimmi20, line 16: I dont like the idea of having a constructor inside a trait.

chappy84: I did consider whether it was a good idea, but a few things made me decide to put it here:

- The self contained nature of the trait, and code deduplication. Without this, everywhere using this trait would have to setup the `propertyHolder` class property, as the required `PropertyHolder` type. With this it just works by being used in classes.
- With the self contained nature, it also makes the trait functional and testable. The trait shouldn't need to rely on structure in classes using it to be functional / testable, only the defined dependency injection required by the constructor.
- The empty nature of the PHP8 constructor, with it's argument. This means there's no code to duplicate / no reason to try to call `parent::__construct` (a primary reason not to include a constructor in a trait). If engineers want to overwrite the constructor, it's a pretty simple process to replicate the functionality anyway, which they would with simply adding an argument of `private PropertyHolder $propertyHolder`. If it were a parent class instead, they'd need a constructor argument of `PropertyHolder $propertyHolder` on the child, and this would need to be passed to the `parent::__construct` call.

asgrim: Hmm, I need to look into this in more detail, but don't currently have the time; in general, introducing any `trait` at all is likely an antipattern. I need to review properly when I have some time to see exactly what you're trying to do here, and whether a better solution can be in place.

chappy84: Full disclosure, I did foresee a better solution, but it breaks BC. These traits would be classes in their own right instead, and the original filter classes where the trait versions are used should be taken away. I say this down to the single responsibility principle, and as mentioned above, there not being any proper link between Divisions / Sections, & Properties.

As there's noting in the v7 milestone, I was trying to avoid breaking BC, and find a half way house between the existing code, and the ultimate destination of completely separate classes. I saw that as a way of getting this functionality in sooner, but equally honestly, should have raised a discussion on the subject instead, to see what the requirements are for a new major version.

mimmi20: Is it possible to an CompositeFilter (or maybe an other name) like that as short term solution?

```php
/**
 * with this filter it is possible to create a custom browscap file
 */
class CompositeFilter implements FilterInterface
{
    /**
     * @throws void
     */
    public function __construct(private FilterInterface $propertyFilter, private FilterInterface $sectionFilter)
    {
    }

    /**
     * returns the Type of the filter
     *
     * @throws void
     */
    public function getType(): string
    {
        return 'composite-filter';
    }

    /**
     * checks if a division should be in the output
     *
     * @throws void
     */
    public function isOutput(Division $division): bool
    {
        return $this->sectionFilter->isOutput($division);
    }

    /**
     * checks if a section should be in the output
     *
     * @param bool[] $section
     *
     * @throws void
     */
    public function isOutputSection(array $section): bool
    {
        return $this->sectionFilter->isOutputSection($section);
    }

    /**
     * checks if a property should be in the output
     *
     * @throws void
     */
    public function isOutputProperty(string $property, WriterInterface $writer): bool
    {
        return $this->propertyFilter->isOutputProperty($property, $writer);
    }
}
```

chappy84: This is very similar to the `CustomFieldStandardFilter` class in my original PR description, where I was trying to demonstrate what people required to get round this current bug. It feels very much like an anti-pattern itself, injecting one filter into another, to try and sort a bug, and one I was trying to avoid when creating the traits. I'd have thought filters should be combined via the `FilterCollection` class instead, but then that's what helped to identify this bug. 🤷‍♂️

chappy84:
#[626bde5](https://github.com/chappy84/browscap/commit/626bde5e158838cdea963005cbe23227d583bf94) is an example of the traits I created being converted fully to individual classes, and being used as individually injected filters to the writers. This obviously breaks BC, but in-part provides an example of one way of solving this, and of what I was saying earlier, providing the separation of concerns with the un-related filters.

PR #2773, opened 2023-01-19
fix validation

mimmi20:
Currently the validation for our resource files does not work.

This PR want to do
- fix the defined json schema used for the validation.
- move the LoggerHelper into the Command/Helper directory because this Helper is only used in the Commands
- add some unittests for the Commands and CommandHelpers


asgrim:
Sorry, checking now...

asgrim (approved):

PR #2958, opened 2024-01-08
allow PHP 8.3

mimmi20:


composer.json
```diff
@@ -33,7 +33,7 @@
     "source": "https://github.com/browscap/browscap"
   },
   "require": {
-    "php": ">=8.1.0,<8.3.0",
+    "php": ">=8.1.0,<8.4.0",
```
asgrim, line 36: Note: we only have to keep 8.1 support until I can get to browscap/browscap-site#354 :+1:

asgrim (approved):

